Issue #28, opened 2020-06-26
Export of results


rneher:
The current export returns a json structure which contains the results, along with a bunch of other things (like processing status, redundant clade definitions, etc). Most users will want to export a file they can view, share, and analyze in Excel or similar. 

I think a tsv table with the following columns would address most needs:
```
sequenceName
clade
alignmentStart
alignmentEnd
numberOfMutations
numberOfNonACGTN
numberOfN
mutations: C242T,C14408T,A24304G,..
deletions: 567-572,23304-20423,...
insertions: 678:ACG,23565:GTGTCG,...
missing: 16121-16525,27343-27565,...
QC-Fail/Pass
QC-flags: 'too many mutations','SNP clusters'
```
Some of the fields are themselves arrays (like the mutations) and hence a table is not ideal. But if we provide json export as well, I think we address most needs.

ivan-aksamentov:
TODO:
 - [ ] remove redundant items from JSON export
 - [ ] implement TSV export as specified above

ivan-aksamentov:
Resolved in #51

PR #51, opened 2020-06-30
Export

ivan-aksamentov:
Implements CSV and JSON export, selectable with a dropdown button.
Only specified subset of results is being exported.


ivan-aksamentov:
Resolves #28

PR #81, opened 2020-07-03
feat: attempt to switch table to react-table + react-window

### LINKED ISSUE #25: Make table sortable and filterable
Allow for sorting and filtering table by sequence names, mutations, possibly other properties.

Possible implementation: react-table
https://github.com/tannerlinsley/react-table

### LINKED ISSUE #35: Put results table in a vertically-scrollable container
Put results table in a fixed-height container such that the table is scrolled separately, without the entire page being scrolled.

Currently, where there are many sequences there might be a lot of scrolling to do just to reach the "Back" button or any other controls on top. Same for Gene Map

This wall make all of the controls accessible at all times. In particular, the Gene Map will always be visible at the bottom (outside scrollable container) and user can scroll a particular sequence closer to it.

Problem: the page will not have scrolling, but we don't necessarily want to waste precious space showing the footer. He have to show Vercel logo somewhere as it was a part of the sponsorship deal though. (#41 )

Related: 
 - table virtualization/windowing #40 
 - zoom feature #34 
 - footer #41

### LINKED ISSUE #40: Use virtualization (windowing) when loading table rows
When there are many sequences in the results table, the app becomes very slow.

We have to limit the amount of rows displayed and to only render rows that are currently visible

For, example this can be achieved with 
https://github.com/bvaughn/react-window

Related:

 - scrollable container #35 
 - searchable and filtering #25

ivan-aksamentov:
This is an ongoing work on replacing the plain `<table/>` on results page.

The plan is to:

 - [ ] use react-table for filtering and grouping (resolves #25)
 - [ ] wrap into scrollable container (resolves #35)
 - [ ] virtualize contents with react-window (resolves #40)


ivan-aksamentov:
Abandoning this idea, because react-table coulmn-oriented design does not work well for us. Also, I no longer see any advantages of using it.

Issue #105, opened 2020-07-12
QC: score-based  output


ivan-aksamentov:
We might want to extend the current QC algorithms to not only produce binary Yes/No output, but also provide more fine-grained results.

For example, for every rule we may produce a numeric score.

These scores can then be used to sort rows by severity of issues and to provide user with more fine-grained message text, upon which it will be easier to take an action.

It is preferable for scores for different rules to be numerically comparable for easier sorting and filtering. That means that more severe issues should have higher weights and produce higher "badness" (or lower "goodness") score. We may allow user to tweak these weights for their needs, to adjust desired QC strictness.

Issue #116, opened 2020-07-14
QC: new structure


ivan-aksamentov:
We want to restructure our quality control (QC) system:

 - [x] have a set of independent QC rules
 - [x] each rule has a set of configuration parameters (e.g. thresholds, coefficients, etc.)
 - [x] each rule outputs a quality score (#105) as well as optional rule-specific outputs
 - [x] each rule has a specific way to be rendered in the UI and in the export files (using rule-specific outputs)
 - [x] each sequence can be highlighted with color ranging from pale yellow to bright red, depending on total QC score
 - [x] rules can be configured or disabled independently

Probably:

 - [ ] treat alignment quality and failures as a QC rule, with it's own score (very high weight)

ivan-aksamentov:
This should now be resolved with #120

PR #120, opened 2020-07-15
feat: score- and rule-based QC

ivan-aksamentov:
Resolves: #105 #116
 
 - [x] have a set of independent QC rules
 - [x] each rule has a set of configuration parameters (e.g. thresholds, coefficients, etc.)
 - [x] each rule outputs a quality score (#105) as well as optional rule-specific outputs
 - [x] each rule has a specific way to be rendered in the tooltips
 - [x] each rule has a specific way to be rendered in the table
 - [x] each rule has a specific way to be rendered in the export files (using rule-specific outputs)
 - [x] adjust filtering
 - [x] adjust sorting
 - [x] each sequence highlighted with color ranging from pale yellow to bright red, depending on total QC score
 - [x] rules can be configured or disabled independently
 - [ ] ~treat alignment quality and failures as a QC rule, with it's own score (very high weight)~
 - [x] normalize score for more precise coloring, filtering and sorting
 - [x] use Nextstrain's tree information to make QC rules even more useful


rneher:
I think this is looking pretty good by now. One thing that deserves some thought maybe is how scores are combined and scaled. Currently, each score is normalized such that the threshold is 100, but whether 50 is "half as bad" as 100 is not clear. We could also combine scores quadratically. 3 scores of 35 should probably not be as bad as one score of 105.

PR #130, opened 2020-07-17
feat: add help tips

ivan-aksamentov:
Resolves #130 

tips


rneher:
excellent! only one minor suggestion. In the text for mutations and sequence view, maybe say that these mutations are "relative to Wuhan-Hu-1 (MN908947)". The sequence is available here: https://www.ncbi.nlm.nih.gov/nuccore/MN908947

PR #131, opened 2020-07-20
feat: locate nodes on auspice tree

ivan-aksamentov:
This PR adds a "Tree" page, which renders auspice tree, showing the sequences on a reference tree.

Initial prototype:

 - [x] add a nextstrain build for generating smaller reference tree

 - [x] for every new node (user-provided sequence), find a node in reference tree that is the closest (in terms of divergence), add new node as a child of the reference node

 - [x] import auspice as a library, figure out all the necessary dependencies, build and initialization requirements (e.g. transpilation, redux, )

 - [x] render tree on a separate page

 - [x] navigate to the tree page and back with buttons on the top panel

 - [x] remove auspice widgets that we don't use, prettify the tree page

Additions after the initial prototype:

 - [x] add QC flags into tree ~tooltips~ popups

 - [x] add list of missing nucs and other analysis results data into tree tooltips

 - [x] add coloring for parent nodes

 - [x] splice new entries to the beginning of dropdown entries array

 - [x] attach new nodes to reference nodes only, don't consider new nodes as parents  when building the tree

Further work:

 - [ ] add auspice genome annotation widget


rneher:
why did you add `nextclades.json`? this would in general be the result of the algorithm, right?

ivan-aksamentov:
@rneher 

> why did you add nextclades.json? this would in general be the result of the algorithm, right?

Right. This is temporary, for easier testing. Will remove during cleanup towards the finish.

Issue #135, opened 2020-07-23
Short sequences are flagged with a QC pass


george-githinji:
1. Very short sequences (1 or even 2 nucleotide) with no ambiguous bases  are flagged as a pass. Could implement a filter for sequence length.

2. How are  lead and trailing 5' and 3' UTRs handled? Should N's in these region comprise a fail in QC?

rneher:
the main issue should be solved. but QC for short sequences is not adequately addressed. We are looking into a better organization of the QC procedure, so will hopefully tackle this along with it.

rneher:
I don't view missing 5' and 3' parts as a QC issue. It likely means only a fragment was sequenced. But the QC criteria should be scaled relative to length of the fragment. Let us know how you'd expect the app to behave.

george-githinji:
One approach might be to disregard Ns within the 3' and 5' regions. This implies identifying these regions from a submitted sequence.

PR #136, opened 2020-07-23
Fix/short alignments

rneher:
This PR implements improvements to how we handle genomic fragments, i.e. sequences that are only a fraction of the genome. The app currently treats short sequences as if there are basically the reference genome, as pointed out by @george-githinji in Issue #135 

this proposed changes 
 * don't attempt to align if sequences are short than 100 bases
 * don't call a clade if some of the clade defining mutations are not covered by the alignment
 * indicated unsequenced regions in the sequence view. 

The remaining question is whether the QC thresholds should scale with the length of aligned fragment. For missing data and divergence this seems plausible.


rneher:
the output now looks like this:
image
the sequence with right part missing has no clade called, the two short ones didn't align.

PR #149, opened 2020-08-03
feat: add tree filtering

### LINKED ISSUE #145: Add tree filtering
We want to be able to filter tree nodes. The underlying functionality will be similar to the filtering links in nextstrain.org/ncov (towards the bottom of the page), but we may change the user interface to be more convenient and discoverable.

This might require the upstream modifications to auspice package.

ivan-aksamentov:
Resolves #145


rneher:
this looks great now. I think this is ready to be released.

Issue #119, opened 2020-07-14
Display some of the PCR primers


ivan-aksamentov:
We want to highlight some of the the primers typically used in PCR.
This can also be further extended to highlighting mutations in these regions.

nabito:
I'm also interested in this feature.

Any idea how to implement it? Or where to first look.

ivan-aksamentov:
@nabito Hello,

For this we will need a list of places where the PCR primers typically are in the DNA/RNA of the virus. Then we just need to highlight them in the SequenceView.

We might also loop over all the mutations in a given sequence and check if it falls into a primer fragment. This can be done somewhere in the analysis part of the algorithm. If there are mutations detected in any of the primer fragments, we may report this in the tooltip of this sequence.

The difficulty here is to find and gather PCR primer data and to convert it to the format convenient to use in the web app (e.g. JSON, containing an array of ranges: { begin, end }, or similar).

PR #198, opened 2020-09-16
feat: PCR primers

ivan-aksamentov:
Resolves: #119 

This PR adds PCR primers into consideration.
In particular it:

 - [x] converts original data to JSON format
 - [ ] ~highlights primer ranges in sequence views~
 - [x] detects and reports mutations in primers
 - [x] excludes primer changes when primer contains matching ambiguous nucleotide
 - [x] formats primer changes in export files
 - [x] shows primer changes as the tree node attribute and as a coloring (presence yes/no)


rneher:
This looks great, @ivan-aksamentov!

The one question that remains for me is whether the ambiguous non-ACGT states are handled correctly on reverse complementation. Their position needs to be mirrored and the state changed. As far as I can tell, this would need to happen here:

https://github.com/nextstrain/nextclade/pull/198/files#diff-00979bfe639fb449fb484e300313a75eR120

state 'R' for example could be 'A' and 'G', after complementation it is 'T', 'C' or state 'Y'. Similarly, the position 3 in a reverse primers would be position -3 from the end after complementation.

ivan-aksamentov:
Addressed the above issues:

 - [x] ambiguous nucleotides are not properly complemented
 - [x] fixed primer positions (ranges) in root sequence when primer is reversed

PR #231, opened 2020-10-09
feat(cli): stream fasta input

ivan-aksamentov:
This allows for input fasta file to be streamed, lifting Node.js's limit to 2 GBytes maximum file size.

We need streaming fasta parser for that, so we switched to using `bionode-fasta` instead of our handwritten one.

In order to implement streaming, the analysis itself, or at least the most time-consuming parts should be streamable, that is the operation performed on each element of the stream should be a pure transform, which can be ran on a separate thread, and independent on other elements of the stream.

This was moslty achieved by making `treeFindNearestNodes()` to operate on a single sequence and bubbling up the loop 
https://github.com/nextstrain/nextclade/blob/e0f64374c243e1013c21a0f8989210a45ca0d53a/packages/web/src/algorithms/tree/treeFindNearestNodes.ts#L93-L106

This allowed to move `treeFindNearestNodes()` itself, as well as clade assignment and qc under the same loop iteration, and then moving them all into the same webworker process. So now most of the algorithm is just a `map()` on a stream of sequences (or an array of sequences in case of web app).

The `treeAttachNodes()` step is however still serial and runs after the stream ends (or after the array `map()` in case of web app). This operation (on would-be-shared-object, the tree) prevents it from being parallel:
https://github.com/nextstrain/nextclade/blob/e0f64374c243e1013c21a0f8989210a45ca0d53a/packages/web/src/algorithms/tree/treeAttachNodes.ts#L129-L135

Because `treeFindNearestNodes()` (and resulting matches) is now in the worker process (one process per sequence) and the `treeAttachNodes()` is in the main thread, and the data between them is passed by copy, we loose the pointer stability we previously (ab)used in order to quickly attach the new sequences.

Therefore, we had to introduce an explicit ID for the nodes and then search the tree for a matching id in `treeAttachNodes()`:
https://github.com/nextstrain/nextclade/blob/e0f64374c243e1013c21a0f8989210a45ca0d53a/packages/web/src/algorithms/tree/treeAttachNodes.ts#L203-L219

which is a serious performance degradation, but is still nothing compared to the core analysis.

--

The code in this PR produces the same exact output as previous implementations.

### DIFF
--- a/packages/web/config/webpack/webpack.cli.babel.ts
+++ b/packages/web/config/webpack/webpack.cli.babel.ts
@@ -47,7 +47,7 @@ module.exports = {
   bail: true,
   name: pkg.name,
   target: 'node',
-  devtool: 'cheap-module-source-map',
+  devtool: 'source-map',
   stats: {
     all: false,
     errors: true,
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -95,6 +95,7 @@
     "autoprefixer": "9.8.6",
     "awesomplete": "1.1.5",
     "axios": "0.20.0",
+    "bionode-fasta": "0.5.6",
     "bootstrap": "4.5.2",
     "classnames": "2.2.6",
     "connected-next-router": "3.1.0",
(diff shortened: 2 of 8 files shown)


ivan-aksamentov:
Obsoleted by #242

PR #239, opened 2020-10-29
Add Subclades to 

rneher:
This PR updates the tree to contain the new 20A.EU1 and 20A.EU2 subclades and adds more interesting and recent example sequences. Before merging, we need to figure out how whether we want to add more right away and when we rerun the entire clade assignment in ingest.


ivan-aksamentov:
I also removed `clades.json` and clade markers on Gene Map, as agreed.

In order to use this in ncov-ingest we need first to merge to master, then release the CLI (beta or alpha build on NPM). After we verify that ingest works as intended we can release to general public (both the app and CLI).

PR #244, opened 2020-11-04
fix: ensure isMatch handles ambiguous nucleotides in both reference a…

rneher:
…nd query

the `isMatch` function was previously assuming that the reference doesn't contain ambiguous nucleotides.


ivan-aksamentov:
I factored away the `intersection()` function and added some more tests to varify the new cases.

Issue #252, opened 2020-11-05
"Unknown Error" for all sequences in CLI with geneMap (v0.7.5, v0.8.0-alpha.6)


fmaguire:
Receiving the ```Error: in sequence "Canada/BC_37_0-2/2020": Unknown error. Please note that this sequence
will not be included in the results.``` error for every sequence ingested into the 0.7.5 CLI versions (docker + direct npm install) using the supplied default `geneMap.json` file but same sequence file is successfully processed in older 0.4.0-beta.6 CLI version, via web interface, or when not using the `-g geneMap.json` option.

# Reproduction
Error happening with all input sequences I've tried, all of below were run with the attached single sequence fasta file [error.txt](https://github.com/nextstrain/nextclade/files/5496152/error.txt) (uploaded with .txt suffix to get around gtihub filetype restrictions).

## Invocations resulting in error
- With the npm install of current main release of the CLI tool (0.7.5):

```
>npm install --global @neherlab/nextclade
>nextclade --input-fasta error.fasta --output-json error_fasta.json -g nextclade/packages/web/src/algorithms/defaults/sars-cov-2/geneMap.json

Error: in sequence "Canada/BC_37_0-2/2020": Unknown error. Please note that this sequence will not be included in the results.
```

- Same error with the npm install of latest alpha (0.8.0-alpha.6):

```
>npm install --global @neherlab/nextclade@alpha
>nextclade --input-fasta error.fasta --output-json error_fasta.json -g nextclade/packages/web/src/algorithms/defaults/sars-cov-2/geneMap.json

Error: in sequence "Canada/BC_37_0-2/2020": geneMap.map is not a function. This sequence will be excluded from further analysis.
```

- Same issue with dockerised CLI tool:

```
>docker pull neherlab/nextclade
>docker run -it --rm -u 1000 --volume="$PWD:/seq" neherlab/nextclade nextclade.js --input-fasta /seq/error.fasta --output-json /seq/error_fasta.json -g /seq/nextclade/packages/web/src/algorithms/defaults/sars-cov-2/geneMap.json

Error: in sequence "Canada/BC_37_0-2/2020": Unknown error. Please note that this sequence will not be included in the results.
```

## Invocations that work
- Succeeds with npm install of current beta release of CLI tool (the older v0.4.0-beta.6)

```
>npm install --global @neherlab/nextclade@beta
[...]

PR #253, opened 2020-11-05
Update CLI readme to install latest alpha

fmaguire:
As pointed out in #252 the current beta package in npm is quite old (0.4.0) so minor change to the documentation to point to the alpha version as that will install the latest non-stable package.


ivan-aksamentov:
Thanks!

I think we should also push stable versions to the alpha and beta channels, and beta versions to alpha channel, so that these truly contain the latest available versions for their quality.

Issue #251, opened 2020-11-05
Add automatic detection of virus type


ivan-aksamentov:
If and when the support for multiple visuses is added #213 we might also add an automatic detection of the virus type from sequences. This would conveniently fill the virus type selection dropdown, for instance.

Issue #213, opened 2020-09-24
Extend for other viruses


ivan-aksamentov:
Currently we only handle SARS-CoV-2 virus correctly.

While it makes sense during pandemic, we might want to extend Nextclade to be able to run for different viruses.

Related: #210

ivan-aksamentov:
#214 now allows customizinf CLI for oteher viruses.

We may eventually implement a proper UI for that in the web app as well. Additionally to everything in #210, we might alsoo ad a dropdown with presets for different viruses (it could be automatically filled as per #251)

PR #271, opened 2020-11-13
feat: add advanced mode, allow custom inputs in the web app

### LINKED ISSUE #122: Add possibility to upload data from results page
Currently user has to go back to main page in order to "upload" more data. We want to avoid this unnecessary navigation by allowing to upload data right on the results page.

 - [x] add button which opens dialog with the file uploader widget, same as on main page
 - [x] make sure it plays well with future result persistence feature #30 

Related:
 - #30 
 - #121

### LINKED ISSUE #205: Add ability to provide custom tree and root sequence in the web app
CLI already have this capability, so all that's needed is to add UI elements to accept the files, read the files content and to pass it to the algorithm instead of defaults.

We might have an expandable section on main page for the advanced params and file uploads #210

### LINKED ISSUE #210: Add expandable "advanced" section for additional input files and parameters on main page
We want to add more control to how algorithm runs in the web app. It requires user to specify some additional parameters, overriding the defaults.

These could be:

 - virus configuration

    - virus name (string)

    - reference tree in Auspice JSON v2 format

    - root sequence (.txt file or string) - should be the root of the reference tree

    - ~definition of clades (CSV and/or JSON file and/or a table widget)~ (no longer used)

    - minimum alignment length (number)

    - PCR primers (CSV and/or JSON file and/or a table widget)
   
    - QC settings (JSON file or a form, like in "Settings" dialog)

and more.

All these parameters should be hidden by default in an expandable section of the interface on main page (near main upload box) and they should be optional.

These parameters ~should~ might persist across app reloads (e.g. in Local Storage).
We might also allow uploading all parameters in one go, e.g. as a single JSON file.

Related: #205

ivan-aksamentov:
This PR aims to implement a new expandable "Advanced" section on the main page (issue #210), which would allow users to provide custom configuration #205, such as custom tree and root sequence.

Resolves #210
Resolves #205
Resolves #122

Related and further work:
 - #213
 - #251


rneher:
the ref tree upload has the wrong label.

PR #275, opened 2020-12-02
feat: take gene map from tree json, account for viruses of different length

ivan-aksamentov:
This PR:

 - removes the stadalone default gene map JSON and takes the default gene map from the tree instead. This is useful when user provides a custom tree. Custom gene map can still be provided separately to override the one in the tree.
 - ensures correct usints of divergence in the final tree (hacky, see: https://github.com/nextstrain/nextclade/pull/275/commits/8db8bde0e30a1d3f4531d88bc522f33901beac2b)
 - ensures that genome size is updated when custom root sequence is uploaded. Useful for viruses of different length.


rneher (approved):
this works for sars and flu.

PR #276, opened 2020-12-07
feat: reimplement aminoacid changes extraction, extract deletions

### LINKED ISSUE #274: aminoacid deletions output
Thanks for a great and easy to use software! 

It looks like nextclade does not report aa deletions directly, while nextstrain+auspice does - is this deliberate? I understand that nextclade reports nt deletions as ranges, but (for me at least) it's intuitive that nextclade and nextstrain+auspice gives somewhat similar output, given it's common nextstrain origin? Reporting aa deletions directly seems like a quite important thing to do, in the context of our work its mandatory with aa deletions - so we would have to write an aditional script to fetch this information. But i'm not a virologist so there might be an obvious reason for this (in that case, sorry!).

ivan-aksamentov:
This implements a new approach to aminoacid changes detection and adds aminoacid deletions detection.

Resolves #274


packages/web/src/algorithms/codonTable.ts
```diff
@@ -2,72 +2,74 @@ import { get } from 'lodash'
...
+  'TCG': 'S' as Aminoacid,
+  'TCT': 'S' as Aminoacid,
+  'TGA': '*' as Aminoacid,
+  'TGC': 'C' as Aminoacid,
+  'TGG': 'W' as Aminoacid,
+  'TGT': 'C' as Aminoacid,
+  'TTA': 'L' as Aminoacid,
+  'TTC': 'F' as Aminoacid,
+  'TTG': 'L' as Aminoacid,
+  'TTT': 'F' as Aminoacid,
```
ivan-aksamentov, line 72: The relevant part is 
```js
'---': AMINOACID_GAP,
```
the rest of the diff just adds quotation marks to the keys (a rule of prettier).

packages/web/src/algorithms/getAllAminoAcidChanges.ts
```diff
@@ -1,57 +0,0 @@
...
-  refSequence: string,
-  geneMap: Gene[],
-): SubstitutionsWithAminoacids[] {
-  return mutations.map(({ pos, queryNuc, refNuc }) => ({
-    pos,
-    refNuc,
-    queryNuc,
-    aaSubstitutions: getAminoAcidChanges(pos, queryNuc, refSequence, geneMap),
-  }))
-}
```
ivan-aksamentov, line 57: With the new approach, the aminoacid changes are no longer associated with the nucleotide changes. That is, the `aaSubstitutions` are no longer a subarray of `nucSubstitutions`, but an entirely separate array (along with the added `aaDeletions`)

This breaks:
 - web UI: we report aminoacid changes for every mutation marker on sequence views
 - export: we currently just dump this array of objects
 - tree attachment: we add aminoacid changes to nodes

packages/web/src/algorithms/types.ts
```diff
@@ -62,8 +62,10 @@ export interface AminoacidSubstitution {
   gene: string
 }
 
-export interface SubstitutionsWithAminoacids extends NucleotideSubstitution {
-  aaSubstitutions: AminoacidSubstitution[]
+export interface AminoacidDeletion {
+  refAA: Aminoacid
+  codon: number
+  gene: string
 }
```
ivan-aksamentov, line 69: This is the new data structure for the AA deletions:

For the end user this can be visualized as `${gene}:${refAA}${codon}-` (e.g. `ORF10:D123-`).

packages/web/src/algorithms/types.ts
```diff
@@ -81,7 +83,7 @@ export interface PcrPrimerChange {
   substitutions: NucleotideSubstitution[]
 }
 
-export interface SubstitutionsWithPrimers extends SubstitutionsWithAminoacids {
+export interface SubstitutionsWithPrimers extends NucleotideSubstitution {
   pcrPrimersChanged: PcrPrimer[]
 }
```
ivan-aksamentov, line 88: The resulting substitutions array will no longer contain `aaSubstitutions` which breaks a few things. See the other comment with the description of the breakage.

packages/web/src/io/getGeneMap.ts
```diff
@@ -40,6 +46,21 @@ export function getGeneMap(geneMapJson: GeneMapJson): Gene[] {
...
+      throw new ErrorGeneInvalidRange(
+        `GeneMap: in gene "${name}": expected gene.begin < gene.end, but got ${begin} < ${end}`,
+      )
+    }
+
+    if (!isDivisibleBy(length, 3)) {
+      throw new ErrorGeneInvalidRange(
+        `GeneMap: in gene "${name}": expected gene length to be divisible by 3, but got ${length}`,
+      )
+    }
```
ivan-aksamentov, line 62: This should ensure that we won't go out of bounds when iterating gene sequences with `i += 3` increments.

rneher:
This looks great! But I have found some small inconsistency in how mutations are assigned to branches on the tree:

image

here the parent branch has the same 6base deletion as the two new sequences. Yet, the amino acid deletion is assigned again and the branch tool-tip shows a deletion of 2 bases. I don't immediately see where this goes wrong, but I can dig a little tomorrow.

rneher (approved):
I think this is good to merge. Aminoacid deletions are now correctly called. We currently don't call partial codons. But we can add this later if we want.

PR #295, opened 2021-01-11
Add sequence view pan and zoom

### LINKED ISSUE #34: Implement zoom and horizontal scrolling  for sequence views
I imagine it to be like Google Maps navigation but in 1 dimension.

Need to figure out what's the best user experience here would be, because vertical scrolling is an important part of the table widget (and should take precedence).

We may for example enable zoom when holding <kbd>Ctrl</kbd>

We all know <kbd>Ctrl</kbd>-to-zoom is much more annoying than modifier-less zoom, so alternatively (or additionally to), we might have a button to toggle <kbd>Ctrl</kbd> modifier requirement (enable/disable "zoom mode")

In any case there should also be buttons for zoom in, zoom out, scroll left scroll right which move the viewport to a given amount of pixels.

Implementation should be aware of screen width, mobile clients. OS differences should be take into account (What was that "different" <kbd>Ctrl</kbd> key on Mac again?)

This should not conflict with zoom or other features in browsers or commonly used browser addons.

ivan-aksamentov:
Adds horizontal zoom and pan (horizontal scroll) in sequence view. It is useful, in particular, to see mutation markers that are close together.

Resolves #34

### DIFF
--- a/packages/web/src/components/Results/ResultsTable.tsx
+++ b/packages/web/src/components/Results/ResultsTable.tsx
@@ -1,4 +1,4 @@
-import React, { memo } from 'react'
+import React, { ChangeEvent, memo, useCallback } from 'react'
 
 import { sum } from 'lodash'
 import { connect } from 'react-redux'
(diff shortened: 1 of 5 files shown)


rneher:
To make this useful, I think we need a few other bits

 * the genemap should move/zoom along with the sequence graph
 * the elements should shrink. They are much wider than one base/one codon and often overlap. when zooming, this overlap should decrease. 

I am wondering whether a mouse-drag-rectangle and pan would be useful.

Issue #300, opened 2021-01-14
Link to "clade-defining mutations" is broken


iskandr:
At the bottom of https://clades.nextstrain.org/ under "Clade Assignment" there is a link for clade-defining mutations which points to https://github.com/nextstrain/ncov/blob/master/config/clades.tsv -- that link doesn't work. I'm guessing it should be https://github.com/nextstrain/ncov/blob/master/defaults/clades.tsv instead.

PR #301, opened 2021-01-14
Updated link for clade defining mutations

iskandr:
Link is currently broken, fixes: https://github.com/nextstrain/nextclade/issues/300


ivan-aksamentov:
@iskandr Thanks, good catch!

PR #358, opened 2021-02-19
Major dependency updrade

ivan-aksamentov:
This upgrades almost all of the dependency packages to the latest versions.

Notable are:
 - React is upgraded from 16 to 17 (https://reactjs.org/blog/2020/10/20/react-v17.html)
 - Webpack is upgraded from 4 to 5. This is especially important because Webpack 5 is needed to make wasm + Threads.js to work nicely together.

As in case with any upgrades, this might have unintended consequences and breakage, but so far it looks okay.


rneher:
This is broken for me on chromium and firefox. Just keeps spinning...

image

rneher:
same in Chrome.

PR #374, opened 2021-03-10
Upgrade auspice to 2.23.0

ivan-aksamentov:
This:
 - upgrades auspice from 2.18.2 to 2.23.0
 - adds the new tree filtering functionality from auspice
 - removes the old tree filtering panel
 - colors tree by clade and filters by "node type - new" by default


rneher:
@jameshadfield: am I correct that these things

https://github.com/nextstrain/nextstrain.org/tree/master/auspice-client/customisations

need to go into auspice of nextclade to make it look more like nextstrain.org?

jameshadfield:
@rneher yes -- all the files in that directory are used but only the `config.json` is passed to `auspice build`. [This is the line](https://github.com/nextstrain/nextstrain.org/blob/master/build.sh#L45) where we build auspice during nextstrain.org deploy.

PR #396, opened 2021-04-26
add Conda and Galaxy badge

bgruening:
This is adding two new badges. An Anaconda badge pointing users to a conda package to install nextclade (and nextalign) and a Galaxy badge pointing users to use nextclade on the European Galaxy server. Nextalign is also packaged and available, but one badge is probably enough. There is also for both packages a BioContainer available (Docker/Singularity).

Greetings from Freiburg!

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,24 @@ by Nextstrain team
   </a>
 </p>
 
+<p align="center">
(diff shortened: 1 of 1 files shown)


ivan-aksamentov:
Thanks @bgruening!

bgruening:
Cool, thanks a lot!

PR #420, opened 2021-06-05
data: update tree with new clade designations and alias names for the…

rneher:
… WHO VoC and VoI names


ivan-aksamentov:
Related: https://github.com/nextstrain/ncov/pull/650

PR #421, opened 2021-06-07
Improve "clade" column

ivan-aksamentov:
This attempts to improve readability of the "Clade" colnum in Nextclade web:

 - Increase width of the "Clade" column
 - Left-align clade names in "Clade" column

The longest versioned clade seem to be "20J (Gamma, V3)" currently and I made clade column to be 120px wide, so that even "22W (Omicron, V25)" fits if it ever happen to exist (please don't)

Before:
before

After:
after

Longer:
longer


rneher:
definitely prettier!

PR #425, opened 2021-06-09
Link aminoacid and nucleotide changes by position

ivan-aksamentov:
Here we establish a many-to-many relation between aminoacid changes (substitutions, deletions) and nucleotide changes (substitutions, deletions).

We group changes by positions:
 - nucleotide substitution is related to aminoacid change iff nucleotide position is within the codon nucleotide range of that aminoacid
 - nucleotide deletion is related to aminoacid change iff nucleotide deletion range intersects the codon nucleotide range of that aminoacid

We only state the fact of spatial locality of these changes and we say nothing about the causal relations between nucleotide changes and aminoacid changes, as this can not always be easily deduced from sequences only.

This data is automatically propagated to JSON result output and to the web app.

Here is what it looks like in the app:

In nucleotide change tooltips:

In aminoacid change tooltips:

### DIFF
--- a/packages/nextclade/CMakeLists.txt
+++ b/packages/nextclade/CMakeLists.txt
@@ -36,6 +36,8 @@ set(SOURCE_FILES
   src/analyze/isMatch.h
   src/analyze/isSequenced.cpp
   src/analyze/isSequenced.h
+  src/analyze/linkNucAndAaChangesInPlace.cpp
+  src/analyze/linkNucAndAaChangesInPlace.h
   src/analyze/nucleotide.cpp
   src/analyze/nucleotide.h
   src/io/CsvWriter.cpp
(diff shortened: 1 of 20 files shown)


emmahodcroft:
I think overall these are looking great. My only comment would be on saying "Amino acid changes nearby". From reading this, it would not be clear to me that this means it's an amino acid change in the codon where this nucleotide is (that's when it pops up, right?) - I would think it might be like... 2 codons over. For me, I interpret 'nearby' as meaning 'not here but not far away' - so it would imply to me that it's _not_ the codon this nucleotide is in.

It might be easier to just say "Amino acid at this codon" (doesn't imply why the AA may have changed or not). I know you're trying to be very careful not to attribute cause, but you also don't want to go too far the other way, implying that this codon might not even be associated with the nucleotide in question!

ivan-aksamentov:
I changed "aminoacids nearby" to "affected codons". Nucleotide change tooltips still say "nearby". We might change it later.

Issue #446, opened 2021-06-14
Analysis of fasta files does not work, error: std::bad_alloc


anneva1:
I tried to analyse fasta files in naxtclade but I got the following error: std::bad_alloc
Can you fix this?
![image](https://user-images.githubusercontent.com/85871849/121869587-7ca8d480-cd02-11eb-945f-4c925c7653ff.png)

Anwer to your questions: 

Operating System: windows 10
Browser: google chrome
Is the error reproducible? Does it happen again if you open a new tab, upload the same Fasta? yes
How big is the Fasta file in MB? I add the fasta file in attachment: 476 kb
Kind regards,

Anne

corneliusroemer:
Thank you very much for reporting the error.

Could you please add the following additional information to enable investigation:
- Operating System
- Browser
- Is the error reproducible? Does it happen again if you open a new tab, upload the same Fasta?
- How big is the Fasta file in MB?

ivan-aksamentov:
Hi @anneva1 ,

Nextclade can sometimes use a lot of memory. `std::bad_alloc` means it cannot find enough memory on your computer in order to run. And we definitely need a better error message here!

In order to help I need to know a few things:

 - how many sequences are you trying to analyze?
 - what operating system, browser and browser version are you using?
 - do you use any ad blocking browser extensions?
 - can you tell how much memory (RAM) does your computer have?
 - how much memory (RAM) is there free before you start sequence analysis?


A few things you might try:

 - try to run with less sequences, smaller fasta file
 - free up some memory by closing unused browser tabs and other programs and documents on your computer
 - try to run on another computer, with more memory resources, if available

anneva1:
Dear,

I even tried to add 1 fasta sequence (SARS-CoV-2) and that even didn't work. It also does not work on the PC of my colleague. It worked perfect until last week. When I tried to analyze my sequences on friday (fasta containing 16 sequences for SARS-CoV-2 analysis) it did not work anymore.

Kind regards,

Anne



    [cid:a6294901-3be9-4216-be46-a9a745ba0e25]

Anne Vankeerberghen
Wetenschappelijk medewerker

Labo Moleculaire Biologie
T. 053/72.40.59
Onze-Lieve-Vrouwziekenhuis VZW Campus Aalst-Asse-Ninove
Moorselbaan 164, B-9300 Aalst
www.olvz.be<http://www.olvz.be/>


________________________________
Van: Ivan Aksamentov ***@***.***>
[...]

PR #451, opened 2021-06-16
Improve user experience when Nextclade runs out of memory, attempt to reduce memory consumption in some cases

ivan-aksamentov:
Attempts to improve user experience around issue #446 

I can reproduce it on a Windows VM when restrict memory to 2Gb and set number of cores to 8.

It seems that the root of the problem is low amount of memory per thread. So either of:

 1. increasing amount of memory available
 2. reducing of number of threads

should help

We cannot control (1), but we can advise user to free up some memory. We can control (2), if we can obtain total system memory available and calculate available memory per thread.

Items:
 - [x] Previously Nextclade just displayed "std::bad_alloc", which is the name of the C++ exception class being thrown on out-of-memory condition. Now it displays a more elaborate message, along with some of the options for mitigation (see screenshot)
 - [x] Additionally, the default number of threads detection now tries to detect amount of system memory available, and reduce number of threads, based on some arbitrary minimum memory size per thread. This is not easy and there are no suitable and widespread APIs for that. It seems that this particular implementation only works in Chromium-based browsers.
 - [x] add settings dialog to allow users to set number of threads manually.
 - [x] increment local storage persistence version to reset the remembered number of threads for all users and to force re-detection


corneliusroemer:
What about detecting OS/Browser combination as a fall back if system memory cannot be detected and reduce # of threads if it's Firefox/Chrome?

packages/web/src/components/Error/ErrorPopup.tsx
```diff
@@ -146,6 +159,79 @@ export function AxiosErrorFailed({ url, status, statusText }: { url: string; sta
...
+          <Li>{t('Close unused web browser tabs, other applications and documents, to free up some memory.')}</Li>
+          <Li>
+            {t('Reduce number of processing threads in Nextclade\'s "Settings" dialog.')}
+            <sup>2</sup>
+          </Li>
+          <Li>
+            {t('Reduce the amount of analyzed data. For example, split large .fasta file to multiple smaller ones.')}
+          </Li>
+          <Li>
+            {t('Consider updating your web browser. Nextclade runs best on the latest versions of {{ browserList }}')}
```
corneliusroemer, line 195: What about suggesting `Edge`? If I remember correctly didn't both bug reporters mention that with `Edge` the bug didn't occur?

packages/web/src/components/Error/ErrorPopup.tsx
```diff
@@ -146,6 +159,79 @@ export function AxiosErrorFailed({ url, status, statusText }: { url: string; sta
...
+      <section className="mt-3">
+        <Alert color="danger" fade={false}>
+          {t('Nextclade tried to allocate system memory, but no additional memory could be allocated.')}
+        </Alert>
+      </section>
+
+      <section className="mt-3">
+        <div>
+          {t(
+            'Nextclade algorithm runs on this computer (and not on remote server or in the cloud) and may require large amounts of computational resources to perform calculations. Please make sure that there is enough system memory available for Nextclade to operate.',
```
corneliusroemer, line 177:
```suggestion
            'The Nextclade algorithm runs entirely locally in this Browser (and not on a remote server or in the cloud) and may thus require large amounts of local computational resources to perform calculations. Please make sure that there is enough system memory (RAM) available for Nextclade to operate.',
```

packages/web/src/components/Error/ErrorPopup.tsx
```diff
@@ -197,14 +289,18 @@ export function ErrorPopupDisconnected({ globalError, algorithmErrors, errorDism
...
                 appName: PROJECT_NAME,
               })}
             </div>
             <div>
               <LinkExternal href={URL_GITHUB_ISSUES}>{URL_GITHUB_ISSUES_FRIENDLY}</LinkExternal>
             </div>
-            <div>{t('The developers will be happy to investigate this problem.')}</div>
+            <div>
+              {t(
+                'so that developers could investigate this problem. Please provide as much details as possible about your input data, operating system, browser version, computer configuration. include other details you deem useful for the diagnostics. Share the example sequence data that allows to reproduce the problem, if possible.',
```
corneliusroemer, line 301:
```suggestion
                'so that developers can investigate this problem. Please provide as much details as possible about your input data, operating system, browser version and computer configuration. Include other details you deem useful for diagnostics. Share the example sequence data that allows to reproduce the problem, if possible.',
```

packages/web/src/components/Error/ErrorPopup.tsx
ivan-aksamentov: I just mention **latest versions** of some browsers with known good WebAssembly implementations, that we test periodically.

We don't monitor Edge routinely and it has low market share, so I would not give it any official mention in the app.

The fact that it worker in Edge might have also been a coincidence. For example, user might have closed Chrome with many tabs, freeing memory and that's why Edge worked. But we won't know for sure.

ivan-aksamentov:
@corneliusroemer 

> What about detecting OS/Browser combination as a fall back if system memory cannot be detected and reduce # of threads if it's Firefox/Chrome?

That's a bit too extreme I think. That would punish everyone in the world because of a few users low-memory machines. It's not that much about browser or OS, as about memory availability. All OSes suffer, it's just some become terribly slow and some throw `bad_alloc`.

I think that the changes in the PR should make it more clear why the error occurs and how to mitigate it. Let's test it and discuss in Slack.

corneliusroemer:
I agree now that we know better under which conditions the error occurs. Will test it.

PR #455, opened 2021-06-22
Add frame shift and stop codon QC checks

ivan-aksamentov:
Introduces 2 new QC checks:

 - frame shift check: checks if any of genes have length that is not divisible by 3
 - stop codon check: checks if any of genes have stop codons anywhere except the last codon

Results currently only exposed to results JSON.


ivan-aksamentov:
Added the new QC checks to the UI

ivan-aksamentov:
Exposed the new QC results to CSV/TSV outputs.

ivan-aksamentov:
All checks are now optional, so that if some of the rule config objects are missing, the rule is considered disabled. Nothing should crash.

There is no way currently to know whether the rule configs were omitted on purpose, by mistake or because of the older version of `qc.json` that does not have these checks yet.

PR #462, opened 2021-06-24
Ignore stop codon ORF8:68*

ivan-aksamentov:
Adds `ORF8:68*` to the ignore list of misplaced stop codons for the "Stop codon" QC rule.


rneher:
yes, I think this is useful. This mutation is present in a good chunk of alpha sequences:

image

PR #467, opened 2021-06-28
Data distribution

ivan-aksamentov:
This PR migrates data from git repo to AWS S3.
 
 - [x] Setup temp S3+Cloudfront
 - [x] Add server for local testing
 - [x] Add data curation scripts to index, download and upload datasets to AWS S3.
 - [x] Add data curator's guide: [here](https://github.com/nextstrain/nextclade/tree/feat/data-distribution/packages/datasets)
 - [x] Fetch data in Nextclade Web - "datasets" page
 - [x] Fetch data in Nextclade Web - main page dropdown
 - [x] Add ~"fetch"~ "dataset get" and "dataset list" commands to Nextclade CLI
 - [x] Add `--with-dataset-dir` flag to the main ("run") command, to run with a dataset from a directory containing dataset
 - [x] Document data fetching in CLI help
 - [x] Document data fetching in CLI docs: [here](https://github.com/nextstrain/nextclade/blob/feat/data-distribution/docs/user/datasets.md)
 - [x] Setup definitive S3+Cloudfront
 - [x] Add gzip compression (files larger than 10Mb are not autocompresed by Cloudfront)
 - [x] Add `Content-Encoding` header


corneliusroemer:
The code here seems to lack example files for `settings.json` and `dataset.json` which are necessary as there seems to be no other documentation. At least the documentation just points towards the example files :)

https://github.com/nextstrain/nextclade/blob/3bc1c91e50b7219d068abdb126dfe3e8f6f46c79/packages/datasets/README.md#L95-L102

ivan-aksamentov:
>The code here seems to lack example files for settings.json and dataset.json which are necessary

They'd be useless without the actual data they describe and we are removing all the data from the repo. 

We might think about enumerating all the files (including intermediate files and index) and allowing to download them from AWS via Cloudfront, so that it does not require credentials. But any meaningful contribution would require write access, and so credentials will be needed for that anyways.

I think this problem will simply cease to exist when this is released and all people get credentials. So I would not invest time there.

PR #468, opened 2021-06-29
Add ARTICnetwork nCoV-2019 PCR primers

ivan-aksamentov:
Adds ARTICnetwork nCoV-2019 V4 PCR primers.

Taken from here:
https://github.com/artic-network/artic-ncov2019/blob/b3f2dda5e6d95bc1c5d95a04d4ef37d304479477/primer_schemes/nCoV-2019/V4/SARS-CoV-2.primer.bed

Also adds a python conversion script to convert .bed the file into Nextclade primers.csv format.


rneher:
feedback I got so far was not very enthusiastic. I am afraid that it will just clutter the output too much for little gain.

PR #470, opened 2021-06-30
Avoid drawing empty unsequenced regions, improve range formatting

ivan-aksamentov:
- Don't draw an sequenced marker if a sample is fully sequenced
 - Warn when formatting an empty or inverted range, and return `"empty range"` string instead of an invalid string representation of it (in both, C++ and JS)

Implementation details:
 - refactor, rename the unsequenced range component and split them into 2 components, for the 5' and 3' ends (it used to be one component, copy-pasted from the gap component)
 - simplify range boundary calculation code and move it into components
 - `formatRange()`: add precondition in C++ and `console.warn()` in JS, when `begin >= end`
 - `formatRange()`: return `"empty range"` when `begin >= end`


packages/web/src/components/SequenceView/SequenceMarkerUnsequenced.tsx
```diff
@@ -0,0 +1,114 @@
...
+}: PropsWithChildren<SequenceMarkerProps>) {
+  const [showTooltip, setShowTooltip] = useState(false)
+
+  const x = begin * pixelsPerBase
+  let width = (end - begin) * pixelsPerBase
+  width = Math.max(width, BASE_MIN_WIDTH_PX)
+
+  if (begin >= end) {
+    console.warn(
+      `SequenceMarker: Attmpted to draw an invalid marker for range: \`[${begin}; ${end})\`. This is probably a bug.`,
```
rneher, line 35: spelling...

rneher (approved):
looks good to me

packages/web/src/components/SequenceView/SequenceMarkerUnsequenced.tsx
ivan-aksamentov: Fixed

PR #485, opened 2021-08-03
fix: avoid crash when the `scale` property is missing in a tree coloring

ivan-aksamentov:
Here we remove assumption that the `meta.colorings[].scale` property is present on colorings of `type` `categorical`. We are also checking if the `scale` is an array, before attempting to push new values into it.

This change should make Nextclade more robust to some of the input tree files.

Only the colorings with keys`region`, `country` and `division` are affected.


corneliusroemer:
Tested it with local build and auspice.json including country. Fix seems to work. LGTM.

PR #493, opened 2021-08-09
feat: simplify tree node info dialog

ivan-aksamentov:
This simplifies the dialog to avoid duplicated and poorly formatted information.

 - [x] Remove "QC Flags" from tree node info
 - [x] Don't show "PCR primer changes: None" when "Has PCR primer changes: No". It still shows "PCR primer changes" when they are present.

Note that this affects the output tree JSON.


corneliusroemer (commented):
Cam we hide all PCR Primer information if the provided `primer.csv` is empty? As will be the case for flu, for example.

ivan-aksamentov:
@corneliusroemer 
> hide all PCR Primer information if the provided primer.csv is empty

Too far away in code. I would not bother.

rneher (approved):

PR #505, opened 2021-08-17
use nextstrain-sphinx-theme in nextclade docs

eharkins:
It was already in the environment for docs here but not being specified in the conf.py file. @ivan-aksamentov let me know if you'd like to petition for any of the options you specified which aren't in https://github.com/nextstrain/sphinx-theme to be included in it.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -72,8 +72,7 @@ html_css_files = [
 html_favicon = '_static/favicon.ico'
 
 html_theme_options = {
-    'display_version': False,
-    'logo_only': True,
+    'logo_only': False,
     'collapse_navigation': False,
     'titles_only': True,
 }


ivan-aksamentov:
Oh, thanks. I had no idea. I just copied the options from augur:
https://github.com/nextstrain/augur/blob/5c89570b456b29bc70377592784f6828eaef8840/docs/conf.py#L89-L95

ivan-aksamentov:
Wait, no, the theme is there, just a few lines above. And if options are removed, I no longer see the "+" buttons for collapsible items.
@eharkins are you sure we need to this? :)

eharkins:
Ah thanks Ivan. Sorry I didn't see that the theme was in there already. I think the issue was that the options you copied were from actually from docs.nextstrain.org which has slightly different options than we have for the various subprojects, e.g. augur. Subprojects need to link back to the main docs.nextstrain.org in the navigation bar which this does now that I changed those options to match the other Nextstrain subprojects.

ivan-aksamentov (approved):

PR #508, opened 2021-08-18
Attempt to detect frame shifts (during alignment)

ivan-aksamentov:
This uses a stateful approach for detecting frame shifts during sequence alignment.

The state is being updated when encountering deletions or insertions during the alignment (`add*()` methods). And the anomaly, if any, is recorded when encountering the next substitution (`reset()` method).

The iteration in the original code is happening on nucleotide sequences (both query and ref) and it goes backwards! (so indices and ranges need to be calculated differently to avoid off-by-1 mistakes).

This is just a POC and more tests and scientific thinking needed to make it correct.

Open questions:
 - What is a good internal representation of a frame shift? Is it a range `[begin; end)`? What useful pieces of information we can keep (and possibly display)?
 - We align both nucleotide sequences and peptides (the modified code is shared between these). Do we need to track shifts in both? Are shifts in peptides even a thing?
 - How to tie nucleotide frame shifts and peptides so that it can be beautifully be shown in the UI for both?
 - What difficulties insertion stripping and gap stripping brings to frame shift detection?
 - We may or may not track the absolute index of a reading frame (`frame` member of the class). It can be just a relative number. In this case perhaps better be called `currentShift` or similar.
 - If the `frame` is the real reading frame index, then we perhaps can avoid `isInFrameShift` boolean state and check for `frame % 3` the same as the initial frame. Although modulo arithmetic operation is terribly slow and we will be doing it on every nucleotide (but that's a premature optimization, let's make code correct first and then think about performance).
 - `reset()` on substitutions may be premature. Frame shifts can span beyond that and perhaps we need to wait until the end of the sequence (or a gene?) to record the anomalies instead.
 - backward iteration might not be appropriate for the frame shift detection. In this case we need to either add another forward loop over sequences, or to find an existing forward loop and stick it there.


ivan-aksamentov:
I realized that backwards iteration won't work or at least harder to make work correctly, because we need information about frame from ***previous*** characters in the sequence.

The alternative approach with forward iteration is in https://github.com/nextstrain/nextclade/pull/533

Issue #520, opened 2021-08-27
[Nextclade Web] Remember last selected dataset


ivan-aksamentov:
It would be nice for dataset selector to remember the dataset that was selected.

We could store this information in local storage.

Care needs to be taken to:

 - ensure the remembered information uniquely identifies the dataset (perhaps store the dataset name and ref sequence accession?)

 - ensure fallback behavior when the remembered dataset no longer exist in the index

ivan-aksamentov:
#525 implements this

Issue #522, opened 2021-08-27
[Nextclade Web] Degraded user experience due to automatic start of the analysis


ivan-aksamentov:
In "Simple mode", the analysis starts automatically after the fasta file dropped into the box. It makes sense when SARS-CoV-2 is the only dataset: there is 1 piece of information needed (fasta file), so after it is provided, there is no need to wait for anything else.

However, with introduction of "Nextclade Datasets" feature (#467), Nextclade Web now requires 2 inputs: fasta file and the name of the dataset.

This makes it awkward in some cases:

 - Users can accidentally trigger the analysis on a wrong dataset if they first provide the file and forget to change the dataset in the dataset selector

 - There is no way to stop the analysis (if incorrect dataset was chosen) [#91](https://github.com/nextstrain/nextclade/issues/91)

 - Changing dataset in the selector does not trigger a new run

 - Removing the fasta file and dropping it again is required to trigger the a run (with the correct dataset now)

What changes need to be implemented to improve user experience?

We might add a "Run" button, similar to what we have in "Advanced mode".

A disadvantage is that SARS-CoV-2 researchers (they are our primary audience, and still enjoy preferential treatment) got used to autostart, which was the feature from the very first release, and adding a button may cause confusion.

We might make it configurable, so that users could opt-in to the autostart if they want to. 
This will be especially convenient if Nextclade could remember the last selected dataset (https://github.com/nextstrain/nextclade/issues/520)

The workflow can be changed substantially when the new datasets widget is implemented https://github.com/nextstrain/nextclade/issues/521.

Perhaps there are other possibilities and small improvements we can add?

ivan-aksamentov:
#525 Introduces new dataset selector. Runs don't start automatically by default. There is a toggle to enable automatic runs (the previous behavior). Datasets are now displayed in a scrabble list, and as a separate step, so that it's more discoverable and harder to miss. Last selected dataset is remembered across page reloads.

Issue #521, opened 2021-08-27
[Nextclade Web] Improve dataset selector


ivan-aksamentov:
The dataset selector is currently implemented as a dropdown above the file drop box.

There is very little space available and the text is crammed into multiple lines. Also it might be challenging for users to select datasets as the number of datasets is expected to grow substantially in the future.

We might implement a better dataset selector widget.

I can think of a table, which displays 1 dataset per row. Clicking on a row or on a "Load" button in the row would load the dataset.

We could group rows by pathogen and ref sequence, as well as allow to optionally show older versions. We might display more useful information and even provide links for downloading zip archives of datasets.

The table would go to the left side of Nextclade main page, replacing the "marketing" widget.

Related:
  - #520

ivan-aksamentov:
#525 implements this

corneliusroemer:
Resolved by #525

PR #525, opened 2021-08-30
[Nextclade Web] Better dataset selector

ivan-aksamentov:
This will implement a better dataset selector to improve looks and user experience around selecting the detesets.

 - [x] larger body, shows details of each dataset
 - [x] wizard-like UX
 - [x] run button (no automatic run by default)
 - [x] link to dataset changelog (see [#611](https://github.com/nextstrain/nextclade/pull/611))
 - [x] redesigned, clarified advanced mode
 - [x] remembers selected dataset
 - [x] styling
 - [x] internal logic cleanup
 - [x] ensure correct interaction with input URL params

Resolves: #520 #521 #522


corneliusroemer:
Really nice.

General feedback:
- I didn't realise at first that the white highlighted dataset is how the selection shows. Maybe turn the `select` button into `selected` and grey it out when that dataset is selected.
- Extend what the search filters on: right now the description and not even the friendly name seem to match. If I enter "influenza" I get no hit.
- For consistency, I would maybe not have the run start when clicking "show example" but rather to load the file as if dropped by user.
- Selection loses state when I navigate back (select H3N2, show example, click back, then selection is on SC2 rather than H3N2)
- Run button sometimes takes seconds to change from grey to green, maybe while the datasets are loaded in the background? This is confusing.
- A click on show example when one has loaded one's own sequences doesn't actually load the example but runs the analysis with one's own sequences. So either should "show example" be greyed out in this case, when own sequences are loaded, or show example should overwrite the user's files.

Minor graphical:
- I'd increase "show example font size and maybe make it a button instead of a text link.
- `default` squishes between Reference Tree and From File:
Screenshot 2021-08-30 at 10 58 51
- General width of the container could be larger on large screens, partially alleviating the problem (there's a lot of unused screen real estate):
image

corneliusroemer:
It looks really good now. Tested it and couldn't find a single issue with it!
image

corneliusroemer:
I like! But for my taste, I feel clicking on a dataset should select it immediately, what's the point of requiring user to click next?
image

ivan-aksamentov:
@corneliusroemer Thanks for checking! I need to say that this is work in progress and I am just exploring things.

The "Next" button is easier to implement currently, but also I thought to add "Previous" button, so "Next" will be needed for symmetry. I might also try adding more pages to this wizard, for example for QC config and other options, which we currently don't expose. But if that's not what's needed, then we might also try the one-click solution as you suggested. 

Anyways, for now I just want to make the internals right and to see the big picture.

I will be back to it next week.

corneliusroemer:
This is nice! Maybe change to `Changelog of dataset updates` to make it explicit what we're pointing to.
image

rneher:
Looks great. A few small issues I noticed:

 * advanced mode is now pretty hidden. You have to select a dataset and then do "show dataset files" and modify one by one. One could make this "show/customize dataset file". Very few people will need this. But maybe could have a "custom data set" option in the selector that proceeds with the list of files open?
 * the file symbol for the gff file is a json symbol
image
 * would it be useful/possible to indicate the current state in the above? Like show "Root sequence (`reference.fasta`)" or `pasted Text`. 
 * some of the text fields with links are pretty small. For example the "load example" one. 
 
image

ivan-aksamentov:
> But maybe could have a "custom data set" option in the selector that proceeds with the list of files open?

You mean in the first step of the wizard, in the list there SC2 and all the flus are? It is somewhat tricky to implement with the current internal design and I think it will be pretty difficult to use. Currently the idea of the advanced mode is that you override the dataset's defaults and the rest is taken from the selected dataset. It will not be possible without selecting a concrete base dataset first. i.e. all files will have to be provided. I imagine most people won't have PCR primers and QC configs handy. 

I am going to reintroduce the "Advanced mode" toggle now, and it will hopefully make it more prominent.


> the file symbol for the gff file is a json symbol

Fixed


> would it be useful/possible to indicate the current state in the above? Like show "Root sequence (reference.fasta)" or pasted Text.

This should already be displayed when the file/url/text is provided. Or do you think we need to also display something by default? I am not sure where and what though. There isn't whole lot of space in the widget and it forces sequence box and run button way down on most screens already.

I added a little explainer text at the top to tell users that these boxes override the defaults. The markdown is in the [packages/web/src/components/Main/AdvancedModeExplanation.mdx](https://github.com/nextstrain/nextclade/blob/feat/better-dataset-selector/packages/web/src/components/Main/AdvancedModeExplanation.mdx?plain=1)


> some of the text fields with links are pretty small. For example the "load example" one.

I wanted to highlight the more important components, like the "Run" button, with size and I thought that the "Load example" link is not important. It actually has the normal, default size. But now that both you and Cornelius mentioned that, I will make it bigger.

rneher:
I like the customize!

rneher:
The current state is very nicely displayed after you load a file or paste something. I was mostly thinking of something that indicates "default" before anything has changed. 

But I really think this is very good. already.

rneher:
I think this is good to merge. We can test drive this a little more, but I think this is ready to see the light of day.

PR #543, opened 2021-09-22
fix: translate past first stop codon in nextclade web

ivan-aksamentov:
After #504 when translation past stop codons became optional, Nextalign and Nextclade CLI gained their flags to toggle this option. However, Nextclade Web was incorrectly using the default value, and translation was terminated on the first stop. 

This fixes the option's value so that Nextclade Web translates full peptides as expected.

This also solves the problem of gigantic peptide context tooltips, caused by deletions that appear in peptides context due to padding with `-` after translation stops. This kind of situations is additionally mitigated in #544.


corneliusroemer (approved):
LGTM
Test with frame shift and stop codon sequences and now it's all clean.

PR #544, opened 2021-09-22
feat(web): truncate peptide context and list of changes in tooltips

ivan-aksamentov:
Currently when a large group of consecutive mutations or deletions appear in a peptide, mouse hovering this group displays a tooltip which may not fit to the screen due to both the width of the peptide context table and height of the list of mutations and deletions.

This PR truncates the context table and the lists, both in the middle, keeping the edges, all making sure tooltips are not too large and that they fit to the screen nicely.

Large tooltips previously appeared due to a bug fixed in https://github.com/nextstrain/nextclade/pull/543, but we should still truncate for the cases where large groups of mutations or deletions appear in user-provided sequences due to various reasons.


corneliusroemer (approved):
LGTM
image

PR #545, opened 2021-09-22
fix: remove mentions of Wuhan-Hu-1 in tooltips

ivan-aksamentov:
Now that we have more pathogens, references to SARS-CoV-2 specifics no longer make sense.


ivan-aksamentov:
Additionally, the typo "if" -> "is" is fixed in 8640e944e38abef32b61fb24e2908df1f0e962ce on master

PR #546, opened 2021-09-22
feat(web): limit width of the summary tooltip

ivan-aksamentov:
The tooltip for "Sequence name"  column of the results table is sometimes too wide, making the table columns inside look disconnected and harder to read. This happens when sequence names are very long.

This PR limits the width of the tooltip, such that the table looks pretty regardless of sequence name length. The sequence name is now forced to wrap to multiple lines if too long.


corneliusroemer (approved):
LGTM
image

PR #549, opened 2021-09-24
feat: mask leading/trailing aa deletions with gap character

ivan-aksamentov:


rneher:
I think the gaps around the frame shifts are quite useful. I am in favor or merging this.

PR #564, opened 2021-10-07
Add aminoacid mutations to the tree branches

ivan-aksamentov:
We want AA mutations to be displayed in tree branch tooltips, just like they do in Nextstrain. Currently it says "No mutations".

We will need to gather a map of all mutations for every reference node on the tree and then calculate the set difference of mutations in the sequence and these reference node mutations and attach the result to the tree's branch attributes.

All this is similar to how it's already done for nucleotide mutations, but with quirks relative to amino acids: ref peptides are needed for lookups and mutations need to be grouped per gene.

### Update

Proper implementation of amino acid private mutations search required some effort. I decided to go ahead and clean things up as we go, so the diff ended up being huge. 

Some code organization changes are:

 - Amino acid private mutations code is now deduplicated and both QC and tree placement use the same results. 
 - Private mutation search code is generic and mostly reused for nucleotides and amino acids, with the small exception where we iterate over deletions - structs have different formats there.
 - The substitutions and deletions structs are now generic so that they can be used in generic code.


corneliusroemer:
This seems to work really well already.

AA muts appear on the branches.

One thing that's missing: AA muts don't appear on the tooltips of tips, maybe you can also add this?

image

Issue #595, opened 2021-11-01
macOS binaries for v1.4.2 fail with `dyld[3912]: Library not loaded: @rpath/libtbb.12.dylib`


jameshadfield:
Reported from a user on intel mac OS 10.14 running the latest pre-built Nextalign binaries (v 1.4.2)

```
dyld: Library not loaded: @rpath/libtbb.12.dylib
 Referenced from: /Users/jwel033/Desktop/seqs/nextalign-MacOS-x86_64
 Reason: image not found
Abort trap: 6
```

Have you seen this before @ivan-aksamentov / any tips on how to debug?

corneliusroemer:
Haven't seen but two tips:
1. Update macOS. `10.14 = Mojave` is no longer supported
2. Try `brew upgrade node` (if you haven't got Homebrew installed, get it), source: https://stackoverflow.com/questions/17703510/dyld-library-not-loaded-reason-image-not-found was pretty easy to find just googling the first line of the error message

corneliusroemer:
I can reproduce 😕 

The new macOS Intel binaries do not seem to work. Both nextalign and nextclade fail with the following error:
```
❯ assets/nextclade-MacOS-x86_64 -v
dyld[3912]: Library not loaded: @rpath/libtbb.12.dylib
  Referenced from: /Users/cr/code/nextclade/assets/nextclade-MacOS-x86_64
  Reason: tried: '/usr/local/lib/libtbb.12.dylib' (no such file), '/usr/lib/libtbb.12.dylib' (no such file)
fish: Job 1, 'assets/nextclade-MacOS-x86_64 -v' terminated by signal SIGABRT (Abort)
```

The release assets for 1.4.1 work without problem.

My system details: macOS 12.0.1 Monterey

corneliusroemer:
I can fix it by installing `tbb` through `brew install tbb` but should we add a dependency to nextclade that requires manual installing through `brew`? Probably not a good idea.

```
❯ assets/nextclade-MacOS-x86_64 -v
dyld[10937]: Library not loaded: @rpath/libtbb.12.dylib
  Referenced from: /Users/cr/code/nextclade/assets/nextclade-MacOS-x86_64
  Reason: tried: '/usr/local/lib/libtbb.12.dylib' (no such file), '/usr/lib/libtbb.12.dylib' (no such file)
fish: Job 1, 'assets/nextclade-MacOS-x86_64 -v' terminated by signal SIGABRT (Abort)

[...]

PR #596, opened 2021-11-01
fix: revert: only consider fully static build on linux 2fcb478

ivan-aksamentov:
This reverts https://github.com/nextstrain/nextclade/commit/2fcb4781ccb1a6c24b49661b865eaf1d04a3f899, which wrongly restricted static linking of dependencies to Linux builds only. As dependency dynamic libraries were now required on macOS, but not redistributed along with the executables, Nextclade was crashing on users' machines, in particular trying to find a dynamic library for Intel TBB.

It went unnoticed during dev and on ci, because these libraries are present in the build directory after the build.

This PR changes build such that macOS binaries are also statically linked and should be self-contained again now.

We might want to implement a smoke test in a clean environment with no build artifacts present other than the executables, to verify that they are truly static and self-contained. This is however challenging for macOS builds, as there are no containers and VMs are quite expensive and clunky to deal with.

Resolves https://github.com/nextstrain/nextclade/issues/595


corneliusroemer:
> We might want to implement a smoke test in a clean environment with no build artifacts present other than the executables, to verify that they are truly static and self-contained. This is however challenging for macOS builds, as there are no containers and VMs are quite expensive and clunky to deal with.

What stops us from simply downloading build artefacts and test with a simple Github action on macOS? Surely this shouldn't be difficult and takes at most 1 minute per action which is cheap. What am I missing?

corneliusroemer:
Seems to work on Intel, I'd merge and release. Can't make things worse even if it doesn't on ARM. Have to get an ARM to test there.

corneliusroemer:
I do get this error: 
```
fish: Job 1, './nextclade \…' terminated by signal --in-order \… (…)
fish: Job SIGSEGV, 'Address boundary error' terminated by signal  ()
```

corneliusroemer (approved):
LGTM, tested on Intel Mac and works

ivan-aksamentov:
@corneliusroemer I'll release this fix and then let's address the segmentation fault as a separate issue.

PR #610, opened 2021-11-18
fix: ensure frame shifts in codon coordinates have size of at least 1

ivan-aksamentov:
Small frame shifts are sometimes erroneously reported as empty in codon coordinates, even if not empty in nucleotide coordinates. 

Consider sequence [GenBank:OU168488](https://cov-spectrum.ethz.ch/public/api/v1/sample/fasta?genbankAccession=OU168488), and its gene N, where a frame shift detected as `[914;917)`  in nucleotide alignment relative coordinates. After transformation to nucleotide reference relative coordinates, and then  to codon coordinates (by dividing by 3), it ends up being `[304,304)` , which is an empty range.

```
      coordinates        begin      end      operation (applied to begin and end)

nuc alignment relative*    914      917
                                         add gene start in alignment coordinates (28273)
nuc alignment absolute** 29187    29190
                                         use coordinate map**** to convert to reference coordinates
nuc reference absolute** 29185    29187
                                         subtract gene start in reference coordinates (28273)
nuc reference relative*    912      914
                                         divide by 3
codon***                   304      304

*   - relative to gene start
**  - absolute here means "relative to sequence start"
*** - codons are always counted relative to gene start
****- allows to lookup alignment coordinates from reference coordinates and vice-versa
```

I think that the division by 3 is incorrect way of transforming nucs to codons here. Smaller shifts map to only a portion of a codon and if both begin and end happen to fall to the same codon, the division gives the same number.

~In this PR I try to simply fixup the ranges after the fact (by adding a method `Range::ensureSizeOfAtLeastOne()`). This ensures that frame shift codon ranges are not empty. I am not sure if there's a better way.~

*Update:* The initial implementation has been generalized, based on Richard's comment below:
We want to also count partially shifted codons and as a particular case, small frame shifts that cover only a portion of a codon. 
So for example, now if 4 nucs are covered, it will now be 2 codons, not 1. Similarly, if 2 nucs are covered, this will be 1 codon, up from 0 (which was the initial bug to fix: empty ranges reported).

The meat of the code is here:
https://github.com/nextstrain/nextclade/blob/bd0c3849811deb7246d222feeb351afebce95c3c/packages/nextalign/src/translate/detectFrameShifts.cpp#L199-L222

While at it, I also added a few invariant checks here and there in the frame shift code (these are the macro that will manifest in dev mode if invalid ranges are encountered).

Before:
before

After:
after


packages/nextalign/include/nextalign/nextalign.h
```diff
@@ -38,6 +38,12 @@ struct Range {
   [[nodiscard]] bool contains(int x) const {
     return x >= begin && x < end;
   }
+
+  void ensureSizeOfAtLeastOne() {
+    if (begin == end) {
+      end += 1;
+    }
+  }
```
ivan-aksamentov, line 46: This is where I extend an empty range to be a range of 1

packages/nextalign/src/translate/detectFrameShifts.cpp
```diff
@@ -199,13 +210,15 @@ std::vector<InternalFrameShiftResultWithMask> translateFrameShifts(//
...
-    Range codon = nucRelRef / 3;
+    Range codon = (nucRelRef / 3);
+    codon.ensureSizeOfAtLeastOne();
 
     Range maskNucRelAln = findMask(query, nucRelAln);
     Range maskNucAbsAln = maskNucRelAln + geneStartAln;
     Range maskNucAbsRef = coordMap.alnToRef(maskNucAbsAln);
     Range maskNucRelRef = maskNucAbsRef - geneStartRef;
     Range maskCodon = maskNucRelRef / 3;
+    maskCodon.ensureSizeOfAtLeastOne();
```
ivan-aksamentov, line 222: The new function is called in these 2 places

packages/nextalign/include/nextalign/nextalign.h
ivan-aksamentov, line 46: This is where I extend an empty range to be a range of 1

rneher:
I am wondering whether it would be more sensible to ceil the right bound to have mod 3=0. like
```
[912,912) -> [912, 912) -> [304,304) -> empty
[912,913) -> [912, 915) -> [304,305)
[912,915) -> [912, 915) -> [304,305)
[912,916) -> [912, 918) -> [304,306)
```
This would be achieved by `upper_ceiled = upper + (3-upper%3)%3` (or alternatively `upper_ceiled = upper - upper%3 + (3?upper%3:0)`).
With this, we wouldn't need to special case `begin==end` and the behavior would be more consistent in that the range always includes all codons touched.

ivan-aksamentov:
> ceil the right bound to have mod 3=0

@rneher 
I tied something like this

```cpp
[[nodiscard]] inline Range nucRangeToCodonRange(const Range& range) {
  return Range{
    .begin = range.begin / 3,
    // Make sure the right boundary is aligned to codon boundary
    .end = (range.end + (3 - range.end % 3) % 3),
  };
}
```

but it crashes on out-of-bounds array accesses in `fillRangeInplace<Aminoacid>()`. Can you double-check is my formula looks correct?

Meanwhile I am digging further. Perhaps we simply need to clamp it to the gene size...  Curiously, this wasn't needed for the "if empty add 1" solution though. But it also might be that the nuc positions are off.

ivan-aksamentov:
Okay, Richard's formula worked. I adjusted the test expectations to the fact that we now count partially covered codons, including the case where there are less than 3 nucleotides shifted. The inspection in Nextclade Web gives the same result as before for OU168488, but for some sequences, frame shifts and the X-masked fragments under them, can now be reported longer by 1 codon.

@corneliusroemer Could you please double check that this solves the problem? And if you have a moment, to also verify not only for OU168488 but also for other sequences you've mentioned.

packages/nextalign/include/nextalign/nextalign.h
```diff
@@ -40,6 +40,14 @@ struct Range {
...
 };
 
+[[nodiscard]] inline Range nucRangeToCodonRange(const Range& range) {
+  return Range{
+    .begin = range.begin / 3,
+    // Make sure the right boundary is aligned to codon boundary
+    .end = (range.end + (3 - range.end % 3) % 3) / 3,
+  };
+}
+
```
ivan-aksamentov, line 50: This is  the new conversion function from nuc to codon positions. note that we adjust `end` to the the next multiple of 3 before the division.

rneher:
this looks like it is working fine. The only question I have whether this changes the frameshifts we consider "known" and ok. These mostly run until the end of the gene and this won't (shouldn't) be affected. If these did change, however, we would need to update this in the data set.

rneher (approved):

corneliusroemer:
I checked and this PR doesn't change any of the known frame shift coordinates.

Also, I can confirm that we no longer get any `empty range` frameshifts in all of the 5m GISAID sequences.

This PR solves the problem of `empty range` but it doesn't solve the problem that for GenBank:OU168488 there should never have been a frame shift in the first place. The alignment is perfect - there's no deletion.

We still seem to have single codon frameshifts - something that's confusing me.

ivan-aksamentov:
Cornelius investigated the issue with single-codon frame shifts. We discussed internally and currently don't know how to fix it. It occurs very very rarely. We may think about it some more, but will handle it as a separate issue.

PR #611, opened 2021-11-18
chore(changelog): mention dataset update 2021-11-16

corneliusroemer:


ivan-aksamentov:
We discussed on Slack and this is not going to work, because the app changelog updates are currently tied to releases.

We need a new mechanism for surfacing the dataset changelog. Perhaps fetch the changelog and render the markdown on runtime.

For now I will add a link to the dataset changelog on GitHub to the bottom of the new dataset selector in the web app (#525)

PR #613, opened 2021-11-19
feat: skip seed match for aa and infer alignment params from nuc alignment

### LINKED ISSUE #594: Failure to align to gene when sequence has single nt deletion in first 251nt of gene
First let me thank you for providing your tools to the community.  They are invaluable to us!

We recently re-ran some sequences through the latest version of Nextclade CLI 1.4 (latest web version as well) that were originally run in version 0.14.4.  We noticed that a few sequences were not aligned to Spike and giving the "Unable to align: no seed matches" message.  A few Spike mutations had previously been called with version 0.14.4 and when we dug into the issue, we found that these particular sequences had an insertion in the beginning of Spike.  Using the reference MN908947 sequence, we introduced artificial single indels and found that any indel from MN908947.3:21563-21814 (first 251 NT of Spike) caused the otherwise wildtype Spike not to be aligned.  We found this to be the case with other genes that we tested as well in the intervals seen below:

 M -> MN908947.3:26523-26605
 N -> MN908947.3:28274-28400
 ORF1a -> MN908947.3:266-505
 ORF1b -> MN908947.3:13468-13719
 ORF3a -> MN908947.3:25393-25488
 ORF7a -> MN908947.3:27394-27459
 ORF8 -> MN908947.3:27894-27959
 S -> MN908947.3:21563-21814

Attached is a file with 2 sequences for each gene listed where we introduced a single deletion at the end of this interval and following the interval.   The sequences are otherwise identical to MN908947.3.

[MN908947_artificialDeletions.txt](https://github.com/nextstrain/nextclade/files/7445094/MN908947_artificialDeletions.txt)

### LINKED ISSUE #609: Spurious and wrong deletions in Spike when aligning with Nextalign
Cross-post from https://discussion.nextstrain.org/t/large-novel-spike-deletions-of-unknown-origin-not-reproducible-outside-nextstrain-framework/845

Minimal reproducible example sequences: https://gitlab.com/-/snippets/2206864

image

Original report by Guy Shapira:
> I was alerted by my colleagues that a newly sequenced SARS-CoV-2 sample had multiple novel deletions in the starting amino acids of the Spike protein.
>
> Bracing for another variant of concern, I ran the sequencing data through my own pipeline (bwa-mem2+iVar) and submitted both to nextclade.
>
> Nextclade identified large Spike deletions in both sequences:
>
> However, I could not find a single indication of these deletions outside of the nextclade results.
>
>Both consensus sequences had the Spike start codon fully in-tact.
>
> My variant caller (iVar) found no such deletions, save for the well known ones further down the gene (69-70 and so forth).
>
> Other tools, such as the GISAID CoVsurver and cov-glue, didn’t find anything out of the ordinary either.
>
> The consensus sequences (same sequencing data, different piplines) can be obtained here.
>
> Do you have any idea what happened here?

Response from @rneher:

> Thanks for flagging this. You are right, these deletions aren’t real.
> 
> the reason seems to be that the amino acid alignment is thrown off by the large numbers of N early in the sequence preceding the deletion around position 65 in spike. We’ll try to fix…

ivan-aksamentov:
This is an implementation ~of Richard's idea here: https://github.com/nextstrain/nextclade/issues/609#issuecomment-973891398~ Update: since then we refined the idea quite more.

The aim is to improve peptide alignment, and to hopefully solve problems like #609 #594, which might be due to inferred band width of the banded alignment being too small.

The proposed solution is to infer the band width and mean shift for peptide alignment from nucleotide alignment, instead of running another seed matching pass.

~This branch is based on #612 and includes debug traces.~ (Not important)

Resolves #609 
Resolves #594


packages/nextalign/src/nextalign.cpp
```diff
@@ -40,24 +43,33 @@ NextalignResultInternal nextalignInternal(const NucleotideSequence& query, const
...
+    const auto refSize = safe_cast<int>(alignmentStatus.result->ref.size());
+    const auto refStrippedSize = safe_cast<int>(refStripped.size());
+
+    const int bandWidth = safe_cast<int>(std::max(querySize - queryStrippedSize, refSize - refStrippedSize) / 3 + 3);
+    const int shift = bandWidth / 2;
+
     auto peptidesInternal = translateGenes(alignmentStatus.result->query, alignmentStatus.result->ref, refPeptides,
-      geneMap, gapOpenCloseAA, options);
+      geneMap, gapOpenCloseAA, options, bandWidth, shift);
     concat_move(peptidesInternal.queryPeptides, queryPeptides);
```
ivan-aksamentov, line 68: This is where the calculation of the band width and of the shift is happening. Then these parameters are carried over into peptide alignment.

ivan-aksamentov, line 68: This is where the calculation of the band width and of the shift is happening. Then these parameters are carried over into peptide alignment.

packages/nextalign/src/translate/translateGenes.cpp
```diff
@@ -72,6 +79,70 @@ std::vector<FrameShiftResult> toExternal(const std::vector<InternalFrameShiftRes
...
+  int trailing = len - end;
+  int total = leading + internal + trailing;
+
+  return GapCounts{
+    .leading = leading,
+    .internal = internal,
+    .trailing = trailing,
+    .total = total,
+  };
+}
```
ivan-aksamentov, line 144: This function counts leading, internal and trailing gaps

packages/nextalign/src/translate/translateGenes.cpp
```diff
@@ -115,7 +186,28 @@ PeptidesInternal translateGenes(                               //
...
+        "Note that this gene will not be included in the results of the sequence.",
+        geneName);
+      warnings.inGenes.push_back(GeneWarning{.geneName = geneName, .message = message});
+      continue;
+    }
+
+    // NOTE: `+ 3` here is a magic number to give some additional space
+    const int bandWidth = safe_cast<int>(std::max(queryGapCounts.internal, refGapCounts.internal) / 3 + 3);
+    const int shift = queryGapCounts.leading + bandWidth / 2;
+    debug_trace("Deduced alignment params: bandWidth={:}, shift={:}\n", bandWidth, shift);
```
ivan-aksamentov, line 211: This estimates bandwidth and shift here based on gap counts. We make sure that that empty peptides are discarded early, so that they don't cause problems.

packages/nextalign/src/nextalign.cpp
ivan-aksamentov: This have been changed now (see below)

packages/nextalign/src/translate/translateGenes.cpp
```diff
@@ -115,7 +186,28 @@ PeptidesInternal translateGenes(                               //
...
```
ivan-aksamentov, line 210: This estimates bandwidth and shift here based on gap counts. We make sure that that empty peptides are discarded early, so that they don't cause problems.

rneher (approved):

rneher:
This looks good to me. It should also make nextclade much less noisy. Our most common previous failure mode was failing seed alignment of genes. This is basically no longer a thing now.

corneliusroemer:
Amazing, this successfully resolves both #613 and #594 - successfully tested!

Issue #607, opened 2021-11-15
flexible specification of properties to infer from the attachment node


rneher:
We currently use the phylogenetic placement to assign a clade to each node (and calculate the private mutations). 

In the future, we might want to infer several properties from that attachment node. For example `clade` and `clade_shortname`, or `Nextstrain_clade` and `WHO_name` or `pango_lineage` or similar. It would thus be useful if the data set somehow specified this and Nextclade was dynamically reading this. The output in the table (web and CLI) would then contain these columns.

PR #615, opened 2021-11-23
feat: use dynamic tree node attributes

ivan-aksamentov:
Resolves [#607](https://github.com/nextstrain/nextclade/issues/607)

This is a prototype for the feature requested in https://github.com/nextstrain/nextclade/issues/607

Nextclade now reads the `node_attrs` from the nearest reference tree node and gathers these values in the analysis results. They are written into output files and are displayed in the UI. They can also be later used in the analysis, QC checks etc.

The keys for the `node_attrs` to consider are expected to be on `meta.extensions.nextclade.clade_node_attrs_keys` of the tree JSON.

### Progress

 - [x] Have a hardcoded list of `node_attrs` to gather
 - [x] Figure out where to get the list of `node_attrs` instead of hardcoded
 - [x] Gather `node_attrs`  from the ref tree nodes
 - [x] Copy these `node_attrs`  from nearest node to the new node
 - [x] Emit these `node_attrs`  into results JSON, into a `customNodeAttributes` object property of a result (of a sequence)
 - [x] Attache `node_attrs` for the new nodes  into the ouput tree JSON, just the way they are on the ref nodes
 - [x] Emit these `node_attrs`  into the CSV and TSV, one colum per attribute
 - [x] Figure out how to get keys of `node_attrs` in the web app
 - [x] Display these `node_attrs` in the Nextclade Web results table 
 - [x] Fix: sort table by custom `node_attrs`
 - [x] Fix: ensure all custom columns are present in the beginning of run

### Example snippet to add to the reference tree

Paste this into `.meta` object:

```json
"extensions": {
  "nextclade": {
    "clade_node_attrs_keys": [
      "GISAID_clade",
      "pango_lineage"
    ]
  }
}
```

###  How to test with CLI

CLI works as long as a tree with `meta.extensions.nextclade.clade_node_attrs_keys` is provided. Output JSON, CSV, TSV will contain custom fields.

### How to test web app:

Option 1:

 - Drop the modified tree in the "Reference tree" section in "Advanced mode"

Option 2:

 - Modify the input tree in the latest version of a dataset in the `data/` directory of the datasets repo
 - Rebuild datasets with `./scripts/rebuld`. Docs: [nextclade_data](https://github.com/nextstrain/nextclade_data)
 - Run custom data server as described in [Test datasets locally](https://github.com/nextstrain/nextclade/blob/master/docs/dev/datasets-local.md)
 - Restart Nextclade Web dev server and it should now pick up the tree from the modified dataset


rneher:
this works well for me. It would be nicer if the custom columns appeared after `clade` in the output table, but this isn't crucial IMO.

ivan-aksamentov:
@rneher 

> It would be nicer if the custom columns appeared after clade in the output table, but this isn't crucial IMO.

They should be immediately to the right from "Clade" column, like on the screenshot. How does it look for you?

ivan-aksamentov:
You probably meant tsv/csv file outputs. I now changed the order of columns there so that the dynamic ones are appended after "clade" column.

rneher (approved):

PR #617, opened 2021-11-25
fix: segfault in coordinate mapper due to out-of-bounds array access

ivan-aksamentov:
This was manifesting as a crash in macOS. Like all memory bugs, it was hard to reproduce, but @corneliusroemer eventually found that it was crashing in the `CoordinateMapper` constructor. After some long staring into the code I found an unguarded access to an empty array.

### DIFF
--- a/packages/nextalign/src/translate/mapCoordinates.cpp
+++ b/packages/nextalign/src/translate/mapCoordinates.cpp
@@ -27,8 +27,12 @@ namespace {
     int refPos = 0;
     for (int i = 0; i < alnLength; ++i) {
       if (ref[i] == Nucleotide::GAP) {
-        const auto& prev = revCoordMap.back();
-        revCoordMap.push_back(prev);
+        if (revCoordMap.empty()) {
+          revCoordMap.push_back(0);
+        } else {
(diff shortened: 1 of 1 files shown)


packages/nextalign/src/translate/mapCoordinates.cpp
```diff
@@ -27,8 +27,12 @@ namespace {
...
       if (ref[i] == Nucleotide::GAP) {
-        const auto& prev = revCoordMap.back();
-        revCoordMap.push_back(prev);
+        if (revCoordMap.empty()) {
+          revCoordMap.push_back(0);
+        } else {
+          const auto& prev = revCoordMap.back();
+          revCoordMap.push_back(prev);
+        }
       } else {
```
ivan-aksamentov, line 36: The access to the last item with `std::vector::back()` of the potentially empty vector was unguarded previously:
```cpp
const auto& prev = revCoordMap.back();
```

I added a conditional to push `0` if the array is empty. Tests pass cleanly, so I think it does not change the behavior, except that the first few elements now contain a well-defined value `0` instead of random garbage from heap.

PR #623, opened 2021-11-28
fix: don't add private reversions when query aa is unknown

ivan-aksamentov:
Here we filter out private substitutions and deletions from results, where positions fall into ranges with aminoacid X in query peptide.

In these cases the information is simply missing from the query and in the `processNodeMutations()` function if a substitution or a deletion is missing from query, but present in the node, a reversion is added to the results (a mutation from the node all the way to what was in the ref seq). We chose to remove these,
because we believe that a reversion is a less likely event and most of these are the sequencing errors followed by uncertainty in translation.

You can see the effect with position 417 here, which is no longer present in private substitutions of a branch:

Before:
before

After:
after

### DIFF
--- a/packages/nextclade/src/tree/findPrivateMutations.cpp
+++ b/packages/nextclade/src/tree/findPrivateMutations.cpp
@@ -30,6 +30,7 @@
 #include <fmt/format.h>
 #include <nextclade/nextclade.h>
 
+#include <algorithm>
 #include <map>
 #include <vector>
 
@@ -40,6 +41,7 @@
 #include "../utils/eraseDuplicates.h"
 #include "../utils/filter.h"
 #include "../utils/mapFind.h"
+#include "../utils/range.h"
 
 
 namespace Nextclade {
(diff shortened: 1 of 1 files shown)


packages/nextclade/src/tree/findPrivateMutations.cpp
```diff
@@ -317,8 +332,26 @@ namespace Nextclade {
...
+      // if a substitution or a deletion is missing from query, but present in the node, a reversion is added to the
+      // results (a mutation from the node all the way to what was in the ref seq). We chose to remove these,
+      // because we believe that a reversion is a less likely event and most of these are the sequencing errors
+      // followed by uncertainty in translation.
+      resultForGene.privateSubstitutions = filter(resultForGene.privateSubstitutions,
+        [&unknownAaRanges](const AminoacidSubstitutionSimple& sub) { return !isUnknownAa(sub.pos, unknownAaRanges); });
+
+      resultForGene.privateDeletions = filter(resultForGene.privateDeletions,
+        [&unknownAaRanges](const AminoacidDeletionSimple& del) { return !isUnknownAa(del.pos, unknownAaRanges); });
+
```
ivan-aksamentov, line 353: The filtering calls are here and the implementation for `isUnknownAa()` is at the top of the same file.

ivan-aksamentov, line 353: The filtering calls are here and the implementation for `isUnknownAa()` is at the top of the same file.

ivan-aksamentov, line 353: The filtering calls are here and the implementation for `isUnknownAa()` is at the top of the same file.

rneher:
This looks mostly good, as far as i can tell. But one thing might need changing. Are you checking for the position to be in an unknown range of any gene? Position 100 in `S` certainly should not depend on whether position 100 is known in `N`. I would think you only have to check the gene in which the mutation you are currently looking at is...

ivan-aksamentov:
@rneher Yes, I filter unknowns for the current gene in advance here:
https://github.com/nextstrain/nextclade/blob/fc24ce2416d747e9bfb9ef5d84d0b6a147f3e48f/packages/nextclade/src/tree/findPrivateMutations.cpp#L335-L336
The result of a filter is an array, so still need for that first `any_of()`

rneher (approved):

corneliusroemer (commented):
Why do we not add `unknown` logic to the if in line 220?
```
if (notCoveredYet && isSequenced && !isGap(nodeQueryNuc)) {
```

Otherwise looks good to me. I added a few comments: https://github.com/nextstrain/nextclade/pull/633

PR #624, opened 2021-11-28
fix: avoid potential dereference of nullopt

ivan-aksamentov:
If there is no open range by the end of the loop, then `begin` might be `nullopt`. Then it's illegal to dereference it in the body of this conditional. I add an additional check to see if `begin` contains a value.

And while we are here, why don't we also check if the range is not empty before pushing it, as well as clarify a couple of comments? (The rule of boy scouts: Always leave the campground cleaner than you found it)

### DIFF
--- a/packages/nextclade/src/qc/rulePrivateMutations.cpp
+++ b/packages/nextclade/src/qc/rulePrivateMutations.cpp
@@ -65,8 +65,8 @@ namespace Nextclade {
       }
     }
 
-    // Terminate the last range with the last deletion
-    if (!privateDeletionsSorted.empty()) {
+    // Terminate the last range if any (there is an open range if `begin` is set)
+    if (!privateDeletionsSorted.empty() && begin) {
       const auto end = privateDeletionsSorted.back().pos;
       const auto length = end - *begin;
       deletionRanges.emplace_back(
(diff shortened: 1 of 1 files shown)


packages/nextclade/src/qc/rulePrivateMutations.cpp
```diff
@@ -61,16 +61,18 @@ namespace Nextclade {
...
     }
 
-    // Terminate the last range with the last deletion
-    if (!privateDeletionsSorted.empty()) {
+    // Terminate the last range if any (there is an open range if `begin` is set)
+    if (!privateDeletionsSorted.empty() && begin) {
       const auto end = privateDeletionsSorted.back().pos;
       const auto length = end - *begin;
-      deletionRanges.emplace_back(
-        NucleotideRange{.begin = *begin, .end = end, .length = length, .character = Nucleotide::GAP});
```
ivan-aksamentov, line 73: Before using `begin` , let's check that it's not `nullopt`, to avoid illegal access, which triggers undefined behavior.

packages/nextclade/src/qc/rulePrivateMutations.cpp
```diff
@@ -61,16 +61,18 @@ namespace Nextclade {
...
+      if (length > 0) {
+        deletionRanges.emplace_back(
+          NucleotideRange{.begin = *begin, .end = end, .length = length, .character = Nucleotide::GAP});
+      }
```
ivan-aksamentov, line 75: If the range is empty, don't push it, because empty ranges don't make sense here.

PR #627, opened 2021-11-29
feat: tune params to improve peptide alignment

ivan-aksamentov:
This improves calculation of bandwidth and shift for peptide alignment, previously introduced in https://github.com/nextstrain/nextclade/pull/613

In this version we ensure that the sequences with large insertions (large number of ref gaps) get a large enough band width.

### DIFF
--- a/packages/nextalign/src/align/alignmentParams.cpp
+++ b/packages/nextalign/src/align/alignmentParams.cpp
@@ -62,10 +62,10 @@ GapCounts countGaps(const NucleotideSequence& seq) {
 }
 
 AlignmentParams calculateAaAlignmentParams(const GapCounts& queryGapCounts, const GapCounts& refGapCounts) {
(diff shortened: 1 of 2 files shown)


ivan-aksamentov:
This PR stems from master and causes invariant violation in the test case ` TranslateGenes.DetectsTrailingFrameShift`:https://github.com/nextstrain/nextclade/blob/b4e027516e737d99179682be3e3e2ecb6b174ce0/packages/nextalign/tests/translateGenes.test.cpp#L65-L98

The message is as follows:
```
packages/nextalign/src/align/alignPairwise.cpp:375:
Invariant violated: expected lastIndexByShift[i] > 0, but got:
lastIndexByShift[i] is -1
and
0 is 0
```

I think this case was fixed in https://github.com/nextstrain/nextclade/pull/625, but might still be worth to double check everything, including the test.

PR #637, opened 2021-12-03
fix: off-by-one errors in insertion positions

ivan-aksamentov:
Multiple related defects were identified:

 1. The `pos` field in the internal objects representing insertions was of by one. `1` needed to be subtracted in insertion detection code to adjust for the fact that `pos` represents a position in reference ***\*after\**** which the insertion occurred. But it did not.
 2. The formatting in the results table in Nextclade Web was incorrectly not converting insertion positions to 1-based, which resulted in off-by-one in the opposite direction, "compensating" the mistake
 3. Nextalign erroneously used a duplicated implementation of the formatting function different from the one in (3) that erroneously did not convert position to 1-based, similar to (2), so the mistake in (1) was not manifesting in insertions CSV file and manifesting in JSON results files (which does not use formatting).
 4. Nextclade CLI formatted the insertion position correctly, and the mistake in (1) was manifesting in insertion CSV and JSON results files.

This PR:
  - A. Fixes the off-by-one error in insertion detection code by subtracting 1 from the `pos`. Addressing (1).

 - B. Adds conversion to 1-based position in insertions, by adding 1 in formatting code in Nextclade Web. Addressing (2). The visible result haven't changed due to combination of (A) and (B). 

 - C. Deduplicates insertion formatting function in Nextalign and Nextclade C++ libraries, by taking the version from Nextclade (case (4) above), which is the correct one. So now the CSV output should be consistent and correct. Addressing (3) and (4).

Additionally, this required to perform some refactoring, because of deduplication and because of the fact functions have moved across library boundaries.

To make this transition cleaner, Nextalign CLI now also converts data structures only before they are output to files, in the executable code, instead of returning pre-formatted ("external") structures from the library. This is for symmetry with Nextclade (the now shared formatting functions expect non-formatted, "internal" inputs) and for convenience.

Note that JSON results are not formatted, but just dump the internal representation, with 0-based positions. This is by design. Insertion positions in JSON have now changed due to (A).

### DIFF
--- a/packages/nextalign/include/nextalign/nextalign.h
+++ b/packages/nextalign/include/nextalign/nextalign.h
@@ -1,5 +1,8 @@
 #pragma once
 
+
+#include <algorithm>
+#include <boost/algorithm/string/join.hpp>
 #include <istream>
 #include <map>
 #include <memory>
@@ -285,28 +288,26 @@ inline bool operator==(const InsertionInternal<Letter>& lhs, const InsertionInte
   return lhs.pos == rhs.pos && lhs.ins == rhs.ins && lhs.length == rhs.length;
 }
 
+using NucleotideInsertion = InsertionInternal<Nucleotide>;
+
+
+template<typename Container, typename Formatter, typename Delimiter>
+std::string formatAndJoin(const Container& elements, Formatter formatter, Delimiter delimiter) {
+  std::vector<std::string> formatted;
+  std::transform(elements.cbegin(), elements.cend(), std::back_inserter(formatted), formatter);
+  return boost::algorithm::join(formatted, delimiter);
+}
+
+std::string formatInsertion(const NucleotideInsertion& insertion);
+
+std::string formatInsertions(const std::vector<NucleotideInsertion>& insertions);
+
 struct Insertion {
   int pos;
   int length;
   std::string ins;
 };
 
-struct NextalignResult {
-  std::string ref;
-  std::string query;
-  int alignmentScore;
-  std::vector<Peptide> queryPeptides;
-  std::vector<Insertion> insertions;
-  Warnings warnings;
-};
-
-struct AlgorithmOutput {
-  int index;
-  std::string seqName;
-  bool hasError;
-  NextalignResult result;
-  std::exception_ptr error;
-};
 
 std::map<std::string, RefPeptideInternal> translateGenesRef(//
   const NucleotideSequence& ref,                            //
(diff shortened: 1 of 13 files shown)


packages/nextalign/include/nextalign/nextalign.h
```diff
@@ -285,39 +288,67 @@ inline bool operator==(const InsertionInternal<Letter>& lhs, const InsertionInte
...
+  const NextalignOptions& options);
+
 struct AlgorithmOutput {
   int index;
   std::string seqName;
   bool hasError;
-  NextalignResult result;
+  NextalignResultInternal result;
   std::exception_ptr error;
 };
```
ivan-aksamentov, line 350: Nextalign CLI now uses "internal", unformatted, data structures and formats them before writing to files, as opposed to receiving "external", formatted outputs. This is for symmetry with Nextclade.

packages/nextalign/src/strip/stripInsertions.h
```diff
@@ -34,7 +34,9 @@ inline StripInsertionsResult<Letter> stripInsertions(const Sequence<Letter>& ref
     if (c == Letter::GAP) {
       if (currentInsertion.empty()) {
         currentInsertion = query[i];
-        insertionStart = i;
+        // NOTE: by convention we set position of insertion to be the index of a character that precedes the insertion,
+        // i.e. a position of reference nucleotide *after* which the insertion have happened.
+        insertionStart = i - 1;
       } else {
         currentInsertion += query[i];
       }
```
ivan-aksamentov, line 42: The fix in (A): off-by-one in the detection code

packages/web/src/components/Results/ListOfInsertions.tsx
```diff
@@ -53,7 +53,7 @@ export function ListOfInsertions({ insertions, totalInsertions }: ListOfInsertio
           <tbody>
             {insertions.map(({ pos, ins }) => (
               <tr key={pos}>
-                <td className="text-center">{pos}</td>
+                <td className="text-center">{pos + 1}</td>
                 <td className="text-center">{ins.length}</td>
                 <td className="text-left">{truncateString(ins, INSERTION_MAX_LENGTH)}</td>
               </tr>
```
ivan-aksamentov, line 59: The fix (B): off-by-one in web app

packages/nextalign/src/nextalign.cpp
```diff
@@ -31,6 +31,18 @@ RefPeptide toRefPeptideExternal(const RefPeptideInternal& peptide) {
...
+std::string formatInsertion(const NucleotideInsertion& insertion) {
+  // NOTE: by convention, in bioinformatics, nucleotides are numbered starting from 1, however our arrays are 0-based
+  const auto positionOneBased = insertion.pos + 1;
+  const auto insertedSequence = toString(insertion.ins);
+  return fmt::format("{}:{}", positionOneBased, insertedSequence);
+}
+
+std::string formatInsertions(const std::vector<NucleotideInsertion>& insertions) {
+  return formatAndJoin(insertions, formatInsertion, ";");
+}
```
ivan-aksamentov, line 43: Fix in (C): The deduplicated functions to format insertions for CSV output. Shared between Nextalign, Nextclade CLI and Nextclade Web (through WebAssembly)

rneher (approved):

PR #638, opened 2021-12-04
feat: dynamically adjust width of AA pixels to avoid overly long muta…

rneher:
…tion groups

### DIFF
--- a/packages/web/src/components/SequenceView/PeptideMarkerMutationGroup.tsx
+++ b/packages/web/src/components/SequenceView/PeptideMarkerMutationGroup.tsx
@@ -29,8 +29,7 @@ export function PeptideMarkerMutation({ change, parentGroup, pixelsPerAa, ...res
   const { codonAaRange } = parentGroup
 
   const pos = codon - codonAaRange.begin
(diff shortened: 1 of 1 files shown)


rneher:
This PR changes how long deletions (or other groups of mutations are displayed) Their size is currently fixed and at least a constant. Hence large groups take up more space than they should and this results in situation like the one below where the red mutation is actually at the same position in the two sequences (253)

image

In the lower case, it is part of a group and the group stretches further to the right than it should. This can also potentially obscure mutations. The PR changes this scaling such that the size of groups doesn't increase linearly with the size:

image

ivan-aksamentov:
@rneher In the new version I don't see the mutation at all. I don't know what trade-offs are the most useful here, so I rely on your opinion and experience using Nextclade.

@corneliusroemer What do you think?

ivan-aksamentov:
An additional or an alternative feature we may consider:
[feat: make results and tree pages full width #640](https://github.com/nextstrain/nextclade/pull/640)

corneliusroemer:
Looks good to me. It would be interesting to merge the wide screen PR into this one so we can tune the width accordingly.

PR #640, opened 2021-12-06
feat: make results and tree pages full width

ivan-aksamentov:
This removes limitations on width of the container that is used in `/results` and `/tree` pages of Nextclade Web. This allows for more space to be occupied by useful information, like the sequence views and the phylo tree, at least on larger screens. 

The layout of the results page is quite complex, so there might be some strange side-effects on smaller screens, especially mobile. But I checked on our main target desktop sizes: 720p, 1080p, 1440p and 4k and it looks good so far.


corneliusroemer:
Looks good to me, tested a bit and couldn't find any issues

PR #644, opened 2021-12-07
fix: center sequence view markers

### LINKED ISSUE #631: Visual: Pixel ranges for unknown aa range and aa substitutions don't seem to match
When analysing Omicron sequences, I noticed that the grey bar for `unknown aa range` does not properly match up with the bar for aa substitutions.

The grey area should overlap visually with the blue bar, but it stops before. The unknown range ends on 862, the blue substitution is at 856. So it looks like the mutation is missing when in fact it's unknown.

Root cause could be, that the blue bars _start_ at 856. They should possibly be _centered_ at 856, then the visual difference would be less.

As it is right now, it's potentially misleading to users.

image

image

ivan-aksamentov:
Resolves #631

This attempts to solve an issue of single nuc/aa sequence markers not aligning properly with the range-based nuc/aa markers. See the linked issue for more details.

This PR:
 - moves single markers by 1/2 of ***\*their\**** width to the left, making them centered around their position
 - moves range markers by 1/2 of ***\*minimum width of a marker\**** to the left, so that their start & end nucs/aas are centered

Let's see if it looks better and whether there are any side-effects


rneher (approved):

rneher (changes requested):
for the ranges, this isn't actually correct. The mutation groups can be wider than part of the alignment they correspond to. So the often need shifting more. Basically, they need to be shifted by `(displayWidth - widthInAlignment)/2`

rneher (approved):

rneher:
The new commit now makes sure that markers and ranges are properly centered even when the display width is bigger than the `pixelsPerAA`. This means that a marker can "overhang" both sides of a range (start/end at `^`, marker `X`):
```
xxXXXXXXXXXXxx
....^----^....
```
While correct, i am not 100% sure it is aesthetically an improvement.

PR #645, opened 2021-12-07
feat(cli): read compressed fasta inputs

ivan-aksamentov:
This allows Nextclade CLI and Nextalign to read compressed fasta files passed to `--input-sequences` and  `--input-fasta`. The compression is detected by file extension:

```bash
nextclade --input-fasta=sequences.fasta
nextclade --input-fasta=sequences.fasta.gz
nextclade --input-fasta=sequences.fasta.xz
```

Using compressed inputs might make tools faster in some cases (depending on I/O speed to CPU speed ratio, reducing read I/O bandwidth may increase performance) and to avoid decompression of large input files, which is costly in terms of time and space.

Progress:

 - [x] refactor to allow different pluggable readers in fasta parser
 - [x] implement plain file reader
 - [x] implement .gz file reader
 - [ ] implement .xz file reader
 - [ ] exclude the decompression code from WASM module compilation


corneliusroemer:
Wouldn't it make sense to then also write out compressed outputs? Writing is slower than reading in most cases. So savings should be even higher.

Is it possible to stream in from stdin? That's something else that could be convenient and potentially more versatile than reading compressed files, since the decompression could happen as a separate command, piping into stdin.

ivan-aksamentov:
> Is it possible to stream in from stdin?

Yes, ncov does it:
https://github.com/nextstrain/ncov/blob/feb2c02dc2b3daba062893753654f578f6bffaa7/workflow/snakemake_rules/main_workflow.smk#L101


> Wouldn't it make sense to then also write out compressed outputs?

Perhaps. But that would be a separate feature.

PR #647, opened 2021-12-07
fix: prevent buffer overflow in coord map

ivan-aksamentov:
This accounts for the fact that the ranges are half-one on the right. So dereferencing `end` index overflows the buffer.

### DIFF
--- a/packages/nextalign/src/translate/extractGene.cpp
+++ b/packages/nextalign/src/translate/extractGene.cpp
@@ -79,18 +79,13 @@ ExtractGeneStatus extractGeneQuery(const NucleotideSequenceView& query, const Ge
   precondition_less(gene.start, gene.end);
 
   // Transform gene coordinates according to coordinate map
-  const auto start = coordMap.refToAln(gene.start);
-  // gene.end is the position after the last base of the gene (0-based indexing)
-  // the corresponding base in the query is hence found by coordMap[gene.end-1]
-  // we add 1 to make that end be again after the last base of the gene.
-  // with this addition: length = end - start
-  const auto end = coordMap.refToAln(gene.end - 1) + 1;
-  const auto length = end - start;
-  // Start and end should be within bounds
-  invariant_less(start, query.size());
-  invariant_less_equal(end, query.size());
+  const auto geneAln = coordMap.refToAln(Range{gene.start, gene.end});
 
-  auto result = NucleotideSequence(details::substr(query, start, length));
+  // Start and end should be within bounds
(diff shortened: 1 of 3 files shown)


rneher:
This should fix it. But while we are at it, why not guard against an empty range? That is make the precondition a real condition and return an empty range if it fails?

packages/nextalign/src/translate/mapCoordinates.cpp
```diff
@@ -95,20 +95,24 @@ int CoordinateMapper::refToAln(int refPos) const {
 }
 
 Range CoordinateMapper::alnToRef(const Range& alnRange) const {
-  precondition_less(alnRange.begin, alnRange.end);
+  if (alnRange.begin >= alnRange.end) {
+    return Range{alnRange.begin, alnRange.begin};
+  }
```
ivan-aksamentov, line 100: Added explicit runtime checks for empty ranges

packages/nextalign/src/translate/mapCoordinates.cpp
```diff
@@ -95,20 +95,24 @@ int CoordinateMapper::refToAln(int refPos) const {
...
   };
   postcondition_less(refRange.begin, refRange.end);
   return refRange;
 }
 
 Range CoordinateMapper::refToAln(const Range& refRange) const {
-  precondition_less(refRange.begin, refRange.end);
+  if (refRange.begin >= refRange.end) {
+    return Range{refRange.begin, refRange.begin};
+  }
```
ivan-aksamentov, line 112: Added explicit runtime checks for empty ranges

packages/nextalign/src/translate/extractGene.cpp
```diff
@@ -79,18 +79,13 @@ ExtractGeneStatus extractGeneQuery(const NucleotideSequenceView& query, const Ge
...
+
   // Start and end should be within bounds
-  invariant_less(start, query.size());
-  invariant_less_equal(end, query.size());
+  invariant_less(geneAln.begin, query.size());
+  invariant_less_equal(geneAln.end, query.size());
 
-  auto result = NucleotideSequence(details::substr(query, start, length));
+  auto result = NucleotideSequence(details::substr(query, geneAln.begin, geneAln.length()));
   const auto resultLength = safe_cast<int>(result.size());
```
ivan-aksamentov, line 89: Use range-based `CoordinateMapper::refToAln()` function to avoid extra "-1 +1"

rneher (approved):

PR #648, opened 2021-12-08
perf: speedup array access in match table lookups

ivan-aksamentov:
This PR optimizes the hottest hostpot of the Nextalign, the match table lookup. It:
 - uses our wrapper `at()` function, instead of `std::array::at()`, to avoid internal runtime bounds checks within C++ standard library in release mode, while still including them in debug mode
 - makes the lookup functions inline (and hence moves them into header file), which should hopefully give the compiler a hint to optimize these function calls more aggressively

Both of these in conjunction give a slight visible improvement:
9.1 seconds -> 8.5 seconds on my machine to align 10k sequnces

I'll instrument and benchmark later, as more micro-optimiztions are coming


corneliusroemer:
Seems to work doing a quick test. Shall I do a more thorough test running it on all 6m sequences and looking for any differences in nextalign results?

ivan-aksamentov:
@corneliusroemer Sure, if it's not too hard or time consuming.

There will be many of these micro-optimizations coming.

corneliusroemer:
I see, well it's still quite a manual process, not really CI style.

Either we automate and run on a chunk, say 1% of all sequence, or I run it manually as a test before release.

ivan-aksamentov:
I prepared a staging cli build here:
https://app.circleci.com/pipelines/github/nextstrain/nextclade/2005/workflows/e420fad3-d528-45e0-9c9a-b707c425f86e

corneliusroemer:
By not CI style I meant that I have to manually do quite a few things on the cluster to run the tests ;)

If you have time, we could integrate the snakemake workflow in CircleCI

https://github.com/corneliusroemer/nextclade-full-run

ivan-aksamentov:
I just prepared binaries so that you have something to run, if you decide to do  manual checks.

> integrate the snakemake workflow in CircleCI

Probably not a great idea to run that on CircleCI. Our VMs there have weak 2 cores and 4 Gb of RAM. It's not suited for any compute.

corneliusroemer:
For CI I would suggest running it on a small subset with maybe 10k sequences, that should complete in a few minutes.

But not really necessary if your existing tests are good enough. The full run is then just for reassurance and to catch rare things.

PR #656, opened 2021-12-21
feat: add Cornelius twitter icon to website

corneliusroemer:


corneliusroemer:
@ivan-aksamentov I don't understand why you and me got reordered, I was right before now you are right.

Also, there seems to be a line break in `NeherLab, Biozentrum, University of Basel` now. Maybe the max width is too small?

image

PR #661, opened 2021-12-27
Fix documentation links

victorlin:
Current links for Nextclade CLI and Nextalign CLI docs show the 404 page.

There might be a better way to provide these links, without the hardcoded `en`.


ivan-aksamentov (approved):

ivan-aksamentov:
Thanks Victor @victorlin!
I modified these 2 links and some others to point to stable version of docs, because this is what prod users should be reading.

PR #675, opened 2022-01-07
fix: undefined behavior in tree json acccess

ivan-aksamentov:
**HOTFIX** goes directly to staging and release, to bypass unreleased master. Reason: urgent fix for a 100% guaranteed crash in production Nextclade CLI.

In case where the key is not present the const version of the get function:

```
const json& get(...) const
```
was returning `defaultValue` which is a reference to a temporary, and hence was destroyed after return, thus the returned reference was dangling.

Here I remove the get function, and move conditionals to the caller. I also simplify the chained access with a json pointer.

### DIFF
--- a/packages/nextclade/src/tree/Tree.cpp
+++ b/packages/nextclade/src/tree/Tree.cpp
@@ -54,14 +54,7 @@ namespace Nextclade {
             divergenceUnitsInt)) {}
   };
 
-  const json& get(const json& js, const std::string& key, const json& defaultValue = json::object()) {
-    if (!js.contains(key)) {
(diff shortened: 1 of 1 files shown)


ivan-aksamentov:
I verified locally that the tree works as expected with and without
```
  "meta"
    "extensions": {
      "nextclade": {
        "clade_node_attrs_keys": [
          "GISAID_clade",
          "pango_lineage"
        ]
      }
    },
 },
```
(this is the part of the tree where the crash is occurred)
and that the outputs are identical to the previous release.

Merging and releasing this, because it cannot make any worse.

PR #678, opened 2022-01-07
fix: use safe containers only in dev mode

ivan-aksamentov:
Since introduction of `safe_vector`, `safe_string` and `save_string_view` classes in  https://github.com/nextstrain/nextclade/pull/671 and https://github.com/nextstrain/nextclade/pull/672, we've observed slowdown on master branch compared to the previous release.

Despite the safe containers being thin wrappers around std containers, they might have introduced an unwanted copy or function call, or perhaps overload resolution does not work the same way and less efficient code is generated.

In any case, these containers are only useful during debugging, so let's conditionally compile them only in debug mode, and alias to std containers in release builds.

### DIFF
--- a/packages/common/include/common/safe_string.h
+++ b/packages/common/include/common/safe_string.h
@@ -1,18 +1,21 @@
 #pragma once
 
+#include <string>
+
+#include "safe_string_view.h"
+
+#if !defined(NDEBUG)
+
 #pragma clang diagnostic push
 #pragma ide diagnostic ignored "OCUnusedStructInspection"
 #pragma ide diagnostic ignored "OCUnusedTypeAliasInspection"
 #pragma ide diagnostic ignored "OCUnusedGlobalDeclarationInspection"
 
-
 #include <algorithm>
 #include <memory>
-#include <string>
 
 #include "contract.h"
 #include "copy.h"
-#include "safe_string_view.h"
 
 
 /**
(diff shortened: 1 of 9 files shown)


ivan-aksamentov:
I ran some numbers, for nextalign on 10k random SC2 sequences:

```bash
/usr/bin/time -f 'Time: %E\nMem:  %M KB\n' ./.out/bin/nextalign-Linux-x86_64 --sequences ~/Desktop/gisaid/gisaid-10k.fasta --reference=data/sars-cov-2/reference.fasta --genemap=data/sars-cov-2/genemap.gff --genes=E,M,N,ORF1a,ORF1b,ORF3a,ORF6,ORF7a,ORF7b,ORF8,ORF9b,S --output-dir=tmp/ --output-basename=nextalign --include-reference --silent
```

Results:

```
branch master, commit 5466240 (CLI release 1.8.1)
beore safe containers were introduced
Time: 0:09.97
Mem:  1169508 KB


branch master, commit 5f58180
with safe containers in production
Time: 0:26.99
Mem:  1419680 KB


branch fix/safe-containers-only-in-dev, commit d1b50c2 (this PR)
with conditionally excluded safe containers
Time: 0:09.90
Mem:  1300288 KB
```

Looks like we are back to normal performance.

rneher (approved):

PR #673, opened 2022-01-06
feature: translate minus strand

dnanto:
@ivan-aksamentov 

- #667
- Adding feature to translate the minus strand sequence as defined in the GFF3 file.
- Adds code for complement and reverse complement copied and modified from nextclade
- Adds test for expected translation of minus-strand gene

Test:
```bash
# data
printf ">ref\nATGACTGAATCCACTTAANNNTTAAGTGGATTCAGTCATGTGGATTCAGTCAT\n" > test.ref.fasta
{
  printf ">alt-1\nATGACTGAATCCACTTAANNNTTAAGTGGATTCAGTCATGTGGATTCAGTCAT\n"
  printf ">alt-2\nATGACTGAATCCACTTAANNNTTAAGTGGATTCAGTCATGTGGATTCCATCAT\n"
  printf ">alt-3\nATGACTCATGAATCCACTTAANNNTTACATGGATTCAGTCATGTGGATTCAGTCAT\n"
  printf ">alt-4\nATGACTGAATCCACTTAANNNTTAAGTGGATTCAGTCATGTGGATTCAGTCAT\n"
  printf ">alt-5\nATGACTGAATCCACTTAANNNTTAAGTGGATTCAGTCATGTGGATTCAGTAGTCAT\n"
  printf ">alt-6\nATGACTGAATCCACTTAANNNTTAAGTAGTGGATTCAGTCATGTGGATTCAGTCAT\n"
} > test.seq.fasta
{
  echo . . gene 1 15 . + 0 gene_name=fwd
  echo . . gene 25 39 . - 0 gene_name=rev
} | tr ' ' '\t' > test.gff3

# run this branch
.build/Release-Static/packages/nextalign_cli/nextalign-Linux-x86_64 \
  -j 1 \
  --in-order \
  -i test.seq.fasta \
  -r test.ref.fasta \
  --include-reference \
  --genes=fwd,rev \
  --genemap test.gff3 \
  --output-dir test1 \
  --min-length 33

# run v1.8.0
conda run -n nextalign \
  nextalign \
    -j 1 \
    --in-order \
    -i test.seq.fasta \
    -r test.ref.fasta \
    --include-reference \
    --genes=fwd,rev \
    --genemap test.gff3 \
    --output-dir test2 \
    --min-length 33
```

Compare:
```
find test1 -type f | sort | while read -r ele; do md5sum "$ele" "${ele/test1/test2}";  done
947cac2f4db72ee50a3be20016ace935  test1/test.seq.aligned.fasta
[...]

PR #685, opened 2022-01-10
feat: add sample data for hAdv-A

ivan-aksamentov:
This adds sample data for Nextalign for Human Adenovirus A.

Extracted with modifications from this message by @dnanto 
https://github.com/nextstrain/nextclade/pull/673#issuecomment-1007488135


corneliusroemer:
What am I expected to see? Adenovirus added as an extra dataset? Doesn't look like it for me 😬

ivan-aksamentov:
@corneliusroemer it's a "nextalign" dataset if you will

corneliusroemer:
Should I see it here? Or how would it manifest? Because I can't find anything 👀 
image

ivan-aksamentov:
@corneliusroemer nono, it is not related to Nextclade at all. There are only sequences.fasta, reference.fasta and genemap.gff, so not enough to run Nextclade. 

I only add it to the collection of Nextalign data in the data/ directory of the repo, alongside enterovirus, zika and flus for which we don't have a Nextclade dataset (a tree basicaly) either.

I'd enjoy of course having a tree and calling it a Nextclade dataset, because Adenoviridae are so different from what we currently work on.

This is mostly for dev and curiosity purposes: to have some samples ready available and to periodically check against them, so that we know we are moving towards the goal of making Nextclade supporting many different viruses.

dnanto:
@ivan-aksamentov @corneliusroemer

I'm currently experimenting with this data set, but I believe that it is valid for testing. Though, I want to mention that these are CDS regions that I've aliased as genes, which I think might be necessary for Nextalign compatibility. I figured that one would want to align codons of the CDS regions instead of entire genes when alternative splicing is a feature (for [some viruses](https://viralzone.expasy.org/1943)). So, this is what motivated appending a suffix to disambiguate the entries. If there are two CDS entries with the same name, then nextalign will override the sequence entry. Anyway, I'll work on a branch to see if I can account for this behavior, that is, if it makes sense or isn't too complex...

Also, I do have a goal of making a full nextclade dataset for HAdV. The current paper I'm working on doesn't require it so far, but I see a potential paper for genotyping purposes, at least for HAdV. Either way, I'm glad to contribute future datasets here and would love to host a complete nextstrain build once the current paper is done.

rneher:
I think we would need to put in some thought on how to handle splicing and multipartite transcripts.

rneher:
For long genomes like that, I think we would also want to optimize the bandwidth determination for the alignment and the seedmatching.

PR #689, opened 2022-01-12
feat: split private mutations into 3 groups: reversions, labeled and unlabeled

### LINKED ISSUE #662: ENH: Add QC rule to spot potential contamination/co-infection/recombination
For various reasons, it can happen that a sequence is a patch work from two or more clades.

It might be helpful if Nextclade flagged such sequences through a new QC rule.

The algorithm could be as simple as:
1. For each major clade, count how many of the private mutations occur in it

Say if the sequence is placed as Omicron, but has private mutations 681R, 452L, it would be flagged as potential Delta contamination/co-infection/recombination.

Motivation: There've been quite some issues with Omicron being contaminated by Delta. With our current QC rules, we are not very sensitive to this with up to ~20 private mutations being OK. With a specific contamination QC rule we could already notice something up with just 2 private mutations.

See for example this case where such a QC rule may have helped: https://twitter.com/AdrianaHeguy/status/1477053001676300289?s=20

ivan-aksamentov:
Resolves #662

We want:
 - a more detailed analysis of private mutations
 - add QC rules to evaluate the results of this analysis

We split private mutations into 3 groups:
 - reversions
 - labeled mutations
 - unlabeled mutations

The reversions are sifted away first. These are the mutations that appear in the nearest tree node, but not in the sequence (because it was reverted to an earlier state).

Further, we use the externally precomputed label data ("label map") to decide whether any given non-reversion private mutation is going to be labeled or unlabeled. Labeled mutations then carry the list of the labels that is taken from the "label map".

Currently this 3-way split is done for substitutions and deletions separately, yielding 6 groups. And of course, aminoacid and nucleotide mutations are handled separately as well.

As of right now we are planning only use the code path for nucleotides. For amino acids, the label map will be empty, so the reversions will be sifted away, but all the remaining mutations will just be under unlabeled group.

Items:
 - [x] split out reversions
 - [x] split out labeled, according to a dummy label map
 - [x] add `virus_properties.json` file containing label map, parse it
 - [x] use real label map in the splitting code
 - [x] restore combined private mutations (used in QC, tree placement, and other parts)
 - [x] render lists of split private mutations in the web app, in the tooltips of "Mut" columns
 - [x] prettify list rendering in the web app
 - [x] emit split mutations into CSV and TSV
 - [x] only use positions and query nucs for label matching
 - [x] add JSON fields for counts
 - [x] add CSV colums for counts
 - [x] ~? render counts in web UI~
 - [x] ~? Add new QC rules~
 - [x] Modify "Private mutations" QC rule to use detailed breakdown of private mutations
 - [x] Modify qc.json to contain weights for the detailed breakdown of private mutations
 - [x] ~? New page with a tabular visualization of shared mutations, similar to [CoVariants](https://covariants.org/shared-mutations). Rows: query sequences; columns: labels from the label map; cells: list of mutations badges.~
 - [x] exclude empty arrays for aminoacid private mutations from JSON output
 - [x] add a file picker for `virus_properties.json` in the advanced mode of web app
 - [x] ensure compatibility and/or smooth fallback for older versions of Nextclade
 - [x] ensure correct functioning of datasets that don't have `virus_properties.json`
 - [x] prepare new dataset scontaining `virus_properties.json` and modified `qc.json` See: https://github.com/nextstrain/nextclade_data/pull/22
 - [x] restore unit tests for private mutations
 - [x] remove split for deletions. We decided that we only apply this feature to substitutions for now. We found that splitting deletions have surprising consequences and edge cases (ranges overlapping with labeled mutations etc.). We may or may not return to splitting deletions, depending on scientific needs.


corneliusroemer:
To test, one needs to drop this file: https://gist.github.com/corneliusroemer/9e45b5d0a35a0b963fdf57662bb5d746

In virus properties: 
image

corneliusroemer:
The private mutation QC doesn't seem to work yet. I'm using quite high multipliers for reversions and labeled muts, but not seeing any change in what's called as problematic.

It may be nice to include the counts in the QC tooltip if it's not too hard to implement.
>Private mutation score: 67 
>(Good < 12 < Mediocre < 40 Bad)
>Reverted substitutions: 5
>Labeled substitutions: 2
>Unlabeled substitutions: 5

corneliusroemer (changes requested):
Oh I see this is outdated

corneliusroemer (approved):

rneher (approved):

rneher:
this looks great to me. I haven't gone through every line that changed, but what I saw makes a lot of sense.

rneher:
I do have a question: I don't quite understand how this works:

https://github.com/nextstrain/nextclade/blob/feat/detailed-private-mutations/packages/nextclade/src/tree/findPrivateMutations.cpp#L239

this is meant to find reversion mutations or reversion of deletions. So you are looking for a situation where the `node` has `-` and the sequence equals the reference. But you seem to be checking `!isGap(nodeQueryNuc)` before precluding the possibility of having the node be `-`. 

Similarly, the reference is never a gap:
https://github.com/nextstrain/nextclade/blob/feat/detailed-private-mutations/packages/nextclade/src/tree/findPrivateMutations.cpp#L244

rneher (commented):
the choice of `seqPositionsCovered` is a bit of an unfortunate name, as not-covered positions in sequencing are the ones that end up as `NNNN`.

https://github.com/nextstrain/nextclade/blob/feat/detailed-private-mutations/packages/nextclade/src/tree/findPrivateMutations.cpp#L330

here, you use this as "dealt with already" is I understand correctly.

ivan-aksamentov:
I tried to address Richards points:

 - renamed `seqPositionsCovered` to `seqPositionsMutatedOrDeleted`, to avoid confusion with "not covered" as in bioinformatics slang
 - removed the condition `!isGap(nodeQueryNuc)`. This is a leftover from when the code was only looking for substitutions
 - removed condition that checks if a ref nuc is a gap. Ref sequence cannot have gaps, because we skip them during parsing: https://github.com/nextstrain/nextclade/blob/a8efc6924e2ea61d359dc029e9802b0a9b722611/packages/nextalign/src/io/kseqpp.h#L65
 

Additionally:

The code that detected reversions of deletions was incorrect. Reversions of deletions are not deletions. So we now treat them as substitutions instead. However, they still have to be grouped into ranges for the purposes of the "private mutations" QC check, similarly to other flavors of deletions.

One unpleasant consequence of reversions of deletions being substitutions now is that they sometimes can be numerous. We might need to treat them as ranges everywhere, not only for the purposes of the QC check.

PR #632, opened 2021-12-01
perf: use faster fasta parser implementation

ivan-aksamentov:
This improves runtime performance of Nextalign and Nextclade CLI.

The speedup is achieved by using a faster FASTA parser. As a base I used code from kseqpp from Ali Ghaffaari (`@cartoonist`): https://github.com/cartoonist/kseqpp

We are loosing these features:

 - Sequence names are no longer tracked and deduplicated. It required storing a map of all sequence names and lookups into this map on each new sequence. So a map of 5M items for GISAID. This feature is a  "nice-to-have" and I think we can sacrifice it.
 - Plain text files are not supported by this parser.
 - Hard to test, because it uses lower-level C APIs to read files and parsing logic intertwined with I/O. Old tests are commented out for now.
 - Requires UNIX-specifics (`unistd.h` and `fcntl.h`). But we only ship Linux and MacOS, so this should be fine. If Windows release will be needed, we can use a second implementation there conditionally.

### Benchmarks

Setup:

 - AWS EC2 c5.24xlarge (96 cores @ 3.6 GHz, 192GiB RAM)
 - SSD: gp2 1TB 3000 IOPS
 - Ubuntu 20.04
 - 100k randomly sampled GISAID sequences

Results:

```
|-------------|---------------------|---------------------|---------|
|             |   Wall clock time   |  Throughput, seq/s  |         |
|-------------|-----------|---------|-----------|---------|---------|
|             |   before  |   after |   before  |  after  | speedup |
|-------------|-----------|---------|-----------|---------|---------|
| Nextalign   |   5m 09s  |     51s |      324  |    1933 |   6.05x |
| Nextclade   |   6m 15s  |  3m 47s |      266  |     440 |   1.65x |
|-------------|-----------|---------|-----------|---------|---------|

```

### Scaling

Both tools now scale nicely to a large number of cores.

Before:
before

After:
after

### Try it:

Staging CLI executables are under "Artifacts" tab on the page for each architecture:
https://app.circleci.com/pipelines/github/nextstrain/nextclade/1873/workflows/d8c194a3-6dbe-4f7b-929a-5a2433b1c470

### Repro

Prepare sample data:

```bash
seqkit sample --proportion=0.02 gisaid.fasta | seqkit head --number=100000 -o gisaid-100k.fasta
seqkit stats gisaid-100k.fasta
```

Alias for time measurement:

```bash
alias TIME="/usr/bin/time -f 'Wall clock time:         %E\nMaximum resident memory: %M KB\n'"
[...]

PR #696, opened 2022-01-24
fix: sanitize parsed sequences correctly

ivan-aksamentov:
The new fasta parser introduced in https://github.com/nextstrain/nextclade/pull/632 is not sanitizing sequences correctly.

The check is only performed until the first valid character, and the the remainder is copied as-is. This does not match the expectations of nextalign algorithm, which assumes input sequences don't contain any gaps for example. This was leading to incorrect alignment and incorrect insertions being output (consisting entirely from gaps) by nextalign and nextclade cli. Web is not affected because it uses the old, slower parser.

In this PR I remove the sanitization from the parser and add an additional sanitization between the parser and alignment. This should ensure that sequences are still read as fast as possible in the parser thread (there is only one) and that sequences are properly sanitized before they enter alignment. The sanitation runs on alignment/analysis threads, and therefore is parallel, however a copy of the sequence string is made in the process. Both may have consequences for performance.

On my machine, running nextalign on 10k random sequences

```
reset; /usr/bin/time -f 'Time: %E\nMem:  %M KB\n' ./.out/bin/nextalign-Linux-x86_64 --sequences ../nextclade_external_data/gisaid/gisaid-10k.fasta --reference=data/sars-cov-2/reference.fasta --genemap=data/sars-cov-2/genemap.gff --genes=E,M,N,ORF1a,ORF1b,ORF3a,ORF6,ORF7a,ORF7b,ORF8,ORF9b,S --output-dir=tmp/ --output-basename=nextalign --include-reference --silent
```

completes in roughly the same time and consumes roughly the same memory:

```
Before:
Time: 0:09.58
Mem:  1337396 KB

After:
Time: 0:09.53
Mem:  1479332 KB
```

We need to verify the correctness of outputs.

### DIFF
--- a/packages/nextalign/src/io/kseqpp.h
+++ b/packages/nextalign/src/io/kseqpp.h
@@ -61,10 +61,6 @@ namespace klibpp {
 
     int64_t index = 0;
 
-    inline bool is_char_allowed(char_type c) {
-      return std::isalpha(c) || c == '.' || c == '?' || c == '*';
-    }
-
   public:
     explicit KStream(Reader reader_) : reader(std::move(reader_)) {
       this->buf.resize(DEFAULT_BUFSIZE);
@@ -120,10 +116,9 @@ namespace klibpp {
       }
 
       while ((c = this->getc()) && c != '>') {
-        if (!is_char_allowed(c)) {
+        if (c == '\n') {
           continue;
         }
-        c = static_cast<char_type>(std::toupper(c));
         rec.seq += c;
         this->getuntil(KStream::SEP_LINE, rec.seq, nullptr, true);// read the rest of the line
       }
--- a/packages/nextalign/src/io/sanitizeSequence.cpp
+++ b/packages/nextalign/src/io/sanitizeSequence.cpp
@@ -1,0 +1,24 @@
+#include <nextalign/nextalign.h>
+
+#include <cctype>
+#include <string>
+
+#include "../utils/safe_cast.h"
+
+namespace {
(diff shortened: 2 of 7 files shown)


packages/nextalign/src/io/kseqpp.h
```diff
@@ -120,10 +116,9 @@ namespace klibpp {
...
 
       while ((c = this->getc()) && c != '>') {
-        if (!is_char_allowed(c)) {
+        if (c == '\n') {
           continue;
         }
-        c = static_cast<char_type>(std::toupper(c));
         rec.seq += c;
         this->getuntil(KStream::SEP_LINE, rec.seq, nullptr, true);// read the rest of the line
       }
```
ivan-aksamentov, line 124: The conditional was skipping the invalid characters only in the beginning, and `getuntil()` would copy all characters until the next newline. This is incorrect.
I removed this check entirely and moved sanitization in a separate step.

packages/nextalign/src/io/sanitizeSequence.cpp
```diff
@@ -0,0 +1,24 @@
...
+  std::string output;
+  output.reserve(str.size());
+  for (char c : str) {
+    if (is_char_allowed(c)) {
+      output += safe_cast<char>(std::toupper(c));
+    }
+  }
+  output.shrink_to_fit();
+  return output;
+}
```
ivan-aksamentov, line 24: This is the new sanitization function

rneher (approved):

Issue #698, opened 2022-01-25
BUG: Ambiguous nucleotides cause private mut to be called as reversion


corneliusroemer:
Investigating EPI_ISL_8927255, I noticed that when there are ambiguous nucleotides in the query sequence at sites that are mutated in the nearest reference tree node, we falsely call this as reversions to the reference.

![image](https://user-images.githubusercontent.com/25161793/150979135-56f7dba4-7bf8-4d21-9811-29a8409b5c28.png)

![image](https://user-images.githubusercontent.com/25161793/150979145-2659a003-1c87-4456-a68d-39500ec88a2b.png)

This is an edge case that we didn't seem to have considered when implementing the new feature, though we could have already thought about it back in the days because it's already relevant for raw private mutation counts: does this count as private mutation or not?

In this particular case, it's very useful to know that there's ambiguity at an attachment node, this is potential evidence that there may be contamination at play. Not quite clear how we would want to report this

PR #702, opened 2022-01-25
fix: account for ambiguous nucs when matching private mut labels

ivan-aksamentov:
Tries to address some of the points in https://github.com/nextstrain/nextclade/issues/698

This replaces strict equality of nucleotides with the equivalence when searching for private mutation labels in the label map.

I replaced `operator==(nuc1, nuc1)` call with `isMatch(nuc1, nuc2)` call, which matches both nucleotides against IUPAC table of equivalence. This means that ambiguous nucleotides will be considered equivalent to their corresponding non-ambuguous ones. Nucleotide `N` will be equivalent to any nucleotide, which may or may not be what we need.


corneliusroemer:
I quickly tested and there seems to be no change yet, that's expected, right? The `isMatch` function is still "TODO".

Maybe convert this to a draft PR?

image
image

ivan-aksamentov:
@corneliusroemer I'd expect this to work for nucs. The "TODO" you are referring to is for the aa version of the function. The nuc version of this function is just above it, and is well implemented, I believe.

The situation you described - I think it is by design:
 - first the reversions are sifted away
 - then non-reversions are split to labeled and unlabeled

This PR only affects the second step. i.e. it does not fix the reversions/non- reversions split.

I will look into the fix of the first step.

corneliusroemer:
I see, you're right, by carefully reading it again I see that you only talk about labeled vs unlabeled mutations when in #698 I especially talk about reversions being the issue. Both should be treated the same, where ambiguous are not counted as private mut if it's consistent - since amiguous mtuations are already accounted for in their own QC rule.

Issue #703, opened 2022-01-25
BUG/UX: `[ERROR] Nextclade: Error: File not found: ""` when neither specifying `--input-dataset` nor `input-virus-properties`


corneliusroemer:
First encountered by @sidneymbell (so impacting actual users)

If you update Nextclade to v1.10.0 and run nextclade without the `--input-dataset` flag (which worked in the past) and don't explicitly add `--input-virus-properties` you get a relatively unspecific error:

> [ERROR] Nextclade: Error: File not found: "". Please verify correctness of input flags. If a dataset is used, check that the dataset is not corrupted and is compatible with this version of Nextclade CLI.

Repro:
```
nextclade run \
                  --input-fasta dataset_old/sequences.fasta \
                  --reference dataset_old/reference.fasta \
                  --input-gene-map dataset_old/genemap.gff \
                  --input-tree dataset_old/tree.json \
                  --input-qc-config dataset_old/qc.json \
                  --input-pcr-primers nextclade_dataset/primers.csv --output-dir . \
                  --output-fasta out.fa
```

When preparing for v1.10.0 we thought about people forgetting to update their datasets, but we didn't think about people who do not use `--input-dataset` and thus will have to manually add `--input-virus-properties`.

Since this issue will likely confuse quite a few users, I suggest we make the error more informative, mentioning the file missing explicitly and/or suggesting to use `--input-dataset`.

In addition, we could make the somewhat breaking change clearer in the changelog to make it easier for people to find out they have to add `--input-virus-properties`.

I quickly searched Github to check whether people do not use `--input-dataset` and there's quite a good number of repos. @fritzo has struggled with this, @dpark01 figured it out in the end, and there are more examples, so I think this is quite high priority.

Example code:
https://github.com/epi2me-labs/wf-artic/blob/master/main.nf#L354
https://github.com/broadinstitute/viral-pipelines/commit/97fd3397479366d7867a8fed6ee1a8cb3e0fa157#diff-3afbf685570ea9a80c1e5eae2b71b260d1c9d27bea5d31138224736974887704
https://github.com/broadinstitute/pyro-cov/commit/5a0abe803539343f8cfa0e17742f12c2177dbbf4
https://github.com/theiagen/public_health_viral_genomics/blob/main/tasks/task_taxonID.wdl

ivan-aksamentov:
@corneliusroemer  I forgot to add a check. Here is the fix: https://github.com/nextstrain/nextclade/pull/704.
[...]

PR #704, opened 2022-01-25
fix: improve error message when virus properties flag is missing

ivan-aksamentov:
Resolves: https://github.com/nextstrain/nextclade/issues/703

Adds missing CLI args check for a situation where `--input-dataset` is not provided and neither the `--input-virus-properties`.

Now the error message reads as follows:

> [ERROR] Nextclade: Error: Either `--input-dataset` or `--input-virus-properties` is required. Cannot proceed without virus properties. Note: the new required file `virus_properties.json` was introduced in version 1.10.0. You can obtain it from the latest Nextclade dataset. See `nextclade dataset get --help` and https://docs.nextstrain.org/projects/nextclade for more information

### DIFF
--- a/packages/nextclade_cli/src/io/getInputPaths.h
+++ b/packages/nextclade_cli/src/io/getInputPaths.h
@@ -100,6 +100,15 @@ namespace Nextclade {
         "Either `--input-dataset` or `--input-qc-config` is required. Cannot proceed without QC config.");
     }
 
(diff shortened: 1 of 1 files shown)


corneliusroemer (approved):
Great! This shouldn't break anything so ready for patch release tomorrow once Richard has ok'ed it?

Issue #708, opened 2022-01-27
BUG/Visual: Display positions of single N markers differ slightly from non-Ns in nuc seq view


danrlu:
Nextclade is SO useful. Thank you for building it and always improving it!! 

We noticed a minor issue. We have 2 sequences, 1 has lots of Ns where they are mutations in the other sequence. The bars don't quite line up:

![image](https://user-images.githubusercontent.com/20667188/151392005-b53ffb54-86fc-4a69-b562-94b8256d8971.png)

Take the blue bar for example:
The position of N:
![image](https://user-images.githubusercontent.com/20667188/151392415-aa9dc5c0-c236-46f7-87d4-d79a75cdafdf.png)

The position of blue bar:
![image](https://user-images.githubusercontent.com/20667188/151392492-376d7526-2bfd-49b9-ae47-6a1839ee6ab3.png)

The yellow bar:
![image](https://user-images.githubusercontent.com/20667188/151392597-04b0117f-013d-4367-a8a0-1c9a790f7539.png)

![image](https://user-images.githubusercontent.com/20667188/151392631-9501b1c0-cbac-437c-9a84-af522ffd678c.png)

corneliusroemer:
Thank you very much for reporting this issue. I agree that single Ns should align with single substitution markers.

The root problem is that to make markers visible, they are magnified. We've tried to improve alignment in the past, centering ranges of mutations - but we may have forgotten about Ns, or single Ns.

May be related to either or both of these commits: https://github.com/nextstrain/nextclade/commit/4048a7b6c959c716fffd5ff79ea485bd0a17cfa2 https://github.com/nextstrain/nextclade/commit/e0d3b379ea006e65e5ccdd3082cba59ebda9fef0

rneher:
this is likely due to one of markers being centered and the other left-aligned. We changed mutations to be centered a while back, but might not have handled the missing marker the same way.

ivan-aksamentov:
I attempted to address this and a similar issue with other sequence markers in the PR https://github.com/nextstrain/nextclade/pull/709

@danrlu Could you please check if it looks any better with the fix, by using this preview app:
https://nextclade-git-fix-center-markers-nextstrain.vercel.app/
?

danrlu:
Beautiful! Thank you all!!
![image](https://user-images.githubusercontent.com/20667188/151428200-3c2e77e9-e9e9-480d-a640-d2739f158af6.png)

PR #709, opened 2022-01-27
fix(web): center markers in sequence views

ivan-aksamentov:
Resolves https://github.com/nextstrain/nextclade/issues/631 https://github.com/nextstrain/nextclade/issues/708 
Followup of https://github.com/nextstrain/nextclade/pull/638 https://github.com/nextstrain/nextclade/pull/644

Continues on the adventure of centering various markers in sequence views, started in https://github.com/nextstrain/nextclade/pull/638 and https://github.com/nextstrain/nextclade/pull/644. 

The goal is to make the different markers to be aligned to their positions in the sequence properly, as well as to be aligned between different other types of markers. The challenge is that there's not enough space to display them and that some of the markers are single positions (substitutions), while others are ranges (Ns), and they can overlap, hide each other, and cause all kinds of other visual anomalies.

This time I adjusted markers:
 - for all ranges (peptide fs, nuc fs, aa Xs, nuc Ns, nuc dels, unsequenced nucs)
 - for nucleotide substitutions (they are non-range, 1-nuc markers)

by subtracting  half of a width of the nuc/aa from their left positions, so that they are shifted left by a half a nuc/aa, ensuring they are anchored at the center of the first nuc/aa in the range. It is important to note that the width of a nuc/aa is variable, so the max of pixel-per-item and min-item-width need to be considered as a width.

We need to check how it plays with the AA mutation groups (which are the single-aa markers aggregated into groups), which has a more complex formula for positions (see https://github.com/nextstrain/nextclade/pull/638, https://github.com/nextstrain/nextclade/pull/644).

I believe that the proper fix would be to add  pan and zoom (#295).

### DIFF
--- a/packages/web/src/components/SequenceView/SequenceMarkerMissing.tsx
+++ b/packages/web/src/components/SequenceView/SequenceMarkerMissing.tsx
@@ -24,9 +24,10 @@ export function SequenceMarkerMissingUnmemoed({ seqName, missing, pixelsPerBase,
   const { begin, end } = missing // prettier-ignore
 
   const id = getSafeId('missing-marker', { seqName, ...missing })
-  const x = begin * pixelsPerBase - pixelsPerBase / 2
   let width = (end - begin) * pixelsPerBase
   width = Math.max(width, BASE_MIN_WIDTH_PX)
+  const halfNuc = Math.max(pixelsPerBase, BASE_MIN_WIDTH_PX) / 2 // Anchor on the center of the first nuc
+  const x = begin * pixelsPerBase - halfNuc
 
   const rangeStr = formatRange(begin, end)
 
--- a/packages/web/src/components/SequenceView/SequenceMarkerMutation.tsx
+++ b/packages/web/src/components/SequenceView/SequenceMarkerMutation.tsx
@@ -56,7 +56,8 @@ function SequenceMarkerMutationDisconnected({
 
   const fill = getNucleotideColor(queryNuc)
   const width = Math.max(BASE_MIN_WIDTH_PX, pixelsPerBase)
-  const x = pos * pixelsPerBase - width / 2
+  const halfNuc = Math.max(pixelsPerBase, BASE_MIN_WIDTH_PX) / 2 // Anchor on the center of the first nuc
+  const x = pos * pixelsPerBase - halfNuc
 
   const totalAaChanges = aaSubstitutions.length + aaDeletions.length
 
--- a/packages/web/src/components/SequenceView/SequenceMarkerUnsequenced.tsx
+++ b/packages/web/src/components/SequenceView/SequenceMarkerUnsequenced.tsx
@@ -26,9 +26,10 @@ export const SequenceMarker = memo(function SequenceMarkerImpl({
 }: PropsWithChildren<SequenceMarkerProps>) {
   const [showTooltip, setShowTooltip] = useState(false)
 
-  const x = begin * pixelsPerBase - pixelsPerBase / 2
   let width = (end - begin) * pixelsPerBase
   width = Math.max(width, BASE_MIN_WIDTH_PX)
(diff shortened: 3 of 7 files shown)


packages/web/src/components/SequenceView/SequenceMarkerUnsequenced.tsx
```diff
@@ -26,9 +26,10 @@ export const SequenceMarker = memo(function SequenceMarkerImpl({
 }: PropsWithChildren<SequenceMarkerProps>) {
   const [showTooltip, setShowTooltip] = useState(false)
 
-  const x = begin * pixelsPerBase - pixelsPerBase / 2
   let width = (end - begin) * pixelsPerBase
   width = Math.max(width, BASE_MIN_WIDTH_PX)
+  const halfNuc = Math.max(pixelsPerBase, BASE_MIN_WIDTH_PX) / 2 // Anchor on the center of the first nuc
+  const x = begin * pixelsPerBase - halfNuc
```
ivan-aksamentov, line 32: I take a max of "pixel per nuc" and "min nuc width", which is the width of a single nuc for this particular marker.

Then I take half of it, which gives a half of a nuc.

I subtract this half-nuc-width from the x position of the rectangle representing the range.

Positions in SVG are left-anchored by default. So subtracting half a nuc shifts the rectangle by half a nuc to the left. This way it looks as if it is anchored at the center of the first nuc in the range.

packages/web/src/components/SequenceView/SequenceMarkerMissing.tsx
```diff
@@ -24,9 +24,10 @@ export function SequenceMarkerMissingUnmemoed({ seqName, missing, pixelsPerBase,
   const { begin, end } = missing // prettier-ignore
 
   const id = getSafeId('missing-marker', { seqName, ...missing })
-  const x = begin * pixelsPerBase - pixelsPerBase / 2
   let width = (end - begin) * pixelsPerBase
   width = Math.max(width, BASE_MIN_WIDTH_PX)
+  const halfNuc = Math.max(pixelsPerBase, BASE_MIN_WIDTH_PX) / 2 // Anchor on the center of the first nuc
+  const x = begin * pixelsPerBase - halfNuc
```
ivan-aksamentov, line 30: Let's take the missing range (N nucs) as an example.

I take a max of "pixels per nuc" and "min nuc width", which gives the width of a single nuc for this particular marker in pixels.

Then I take half of this number, which gives a half of a nuc.

I subtract this half-nuc-width from the x position of the rectangle representing the range.

Elements in SVG are anchored on their left edges by default. So subtracting half a nuc shifts the rectangle by half a nuc to the left. This way it looks as if it is anchored at the center of the first nuc in the range.

packages/web/src/components/SequenceView/SequenceMarkerMutation.tsx
```diff
@@ -56,7 +56,8 @@ function SequenceMarkerMutationDisconnected({
 
   const fill = getNucleotideColor(queryNuc)
   const width = Math.max(BASE_MIN_WIDTH_PX, pixelsPerBase)
-  const x = pos * pixelsPerBase - width / 2
+  const halfNuc = Math.max(pixelsPerBase, BASE_MIN_WIDTH_PX) / 2 // Anchor on the center of the first nuc
+  const x = pos * pixelsPerBase - halfNuc
```
ivan-aksamentov, line 60: This is the only non-range marker changed: the nuc substitution marker.

Similarly, I shift it to the left by half a nuc. Note it was already shifted before by (width / 2). But because width is always 1 nuc, I find it more explicit to subtract the half a nuc. Gives the same result, but looks more in line with the rest of markers.

rneher:
The challenge with mutation groups was that we need to center the group, not the first mutations. But I don't think this is necessary here. So I think this is good.

corneliusroemer (approved):
Looks good in the test build for single nucs.

Ranges are still not perfect but that's simply an impossibility due to magnification.

Issue #707, opened 2022-01-26
BUG: Reverted deleted nucs are counted as individual private subs in calculations of SNP cluster rule


corneliusroemer:
Private deletion reversions are counted as individual private mutations in the SNP cluster QC rule. 

So if you have a stretch of 9 deletions plus 2 other private muts within a window of 100, you get to 11 private muts, firing a cluster.

Introduced in version 1.13.0 with new private mut handling.

For sequences: EPI_ISL_8610834 EPI_ISL_8898412

Compare v1.13 (web):
![image](https://user-images.githubusercontent.com/25161793/151201764-3bc34a77-56ae-49b0-8ab7-5b7de4aedf8c.png)

And v1.12.0 (web):
![image](https://user-images.githubusercontent.com/25161793/151201812-dab0ea27-00bd-4cf9-83fc-82a4ce682b2d.png)

These are the non-deletion private mutations:
![image](https://user-images.githubusercontent.com/25161793/151201865-a858b589-68ce-4046-92ea-712a573094d9.png)

(Noticed this behaviour in #706, realised it's a bug so opened new issue)

PR #710, opened 2022-01-27
fix: exclude reversions of dels from consideration in "SNP clusters" QC

ivan-aksamentov:
Resolves https://github.com/nextstrain/nextclade/issues/707
Related to https://github.com/nextstrain/nextclade/pull/689

This tries to address the fact that reversions of deletions are now included into "all private mutations" array and so are processed in the "SNP clusters" QC rule.
Because sometimes reversions of deletions can come in large contiguous chunks, these mutations contribute to the cluster counts, making QC scores for some of the sequences worse. This was unexpected.

In order to mitigate this, it was decided in the team chat to exclude reversions of deletions from the consideration in the "SNP clusters" QC rule.

The technical implementation is to remove the intersection of "all private substotutions" and "reversions of deletions" from "all private substitutions". Implemented rather naively, because these arrays are small and don't worth the hassle of using any more advanced algorithms.

We may or may not to reconsider this. One thing to try could be to treat reversions of deletions as ranges. This would require introducing a concept of ranges of substitutions (similar to the currently existing ranges of deletions) or to treat reversions of deletions as insertions. The QC rule then could be adjusted to count these ranges and/or to assign a lower weight. In any case it is substantially more work and potential breaking changes to file formats, compared to just excluding them. And because the "SNP clusters" rule have somewhat less of importance nowadays, it was decided to go the easy path for now and to focus on other, more important issues.

### DIFF
--- a/packages/nextclade/src/qc/ruleSnpClusters.cpp
+++ b/packages/nextclade/src/qc/ruleSnpClusters.cpp
@@ -1,17 +1,32 @@
 #include "ruleSnpClusters.h"
 
+#include <common/safe_vector.h>
 #include <nextclade/nextclade.h>
 
 #include <algorithm>
 #include <deque>
 #include <optional>
-#include <common/safe_vector.h>
 
+#include "../utils/concat.h"
 #include "../utils/safe_cast.h"
 #include "getQcRuleStatus.h"
 
 
 namespace Nextclade {
+  namespace {
+    // A naive way to erase mutations from a vector of mutations, that are present in another vector of mutations
+    safe_vector<NucleotideSubstitutionSimple> eraseIntersection(const safe_vector<NucleotideSubstitutionSimple>& big,
+      const safe_vector<NucleotideSubstitutionSimple>& small) {
+      safe_vector<NucleotideSubstitutionSimple> result;
+      std::copy_if(big.begin(), big.end(), std::back_inserter(result), [&small](const NucleotideSubstitutionSimple& b) {
+        return small.end() == std::find_if(small.cbegin(), small.cend(),
+                                [&b](const NucleotideSubstitutionSimple& s) { return b.pos == s.pos; });
+      });
+      return result;
+    }
+
+  }// namespace
+
   safe_vector<safe_vector<int>> findSnpClusters(                         //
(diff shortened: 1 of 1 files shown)


ivan-aksamentov:
I don't know why it happens, but sometimes I have an urge to write long descriptions for PRs, even for small changes like this one. I should probably focus on something else instead.

packages/nextclade/src/qc/ruleSnpClusters.cpp
```diff
@@ -72,8 +87,13 @@ namespace Nextclade {
       return {};
     }
 
+    // NOTE: we exclude reversions of deletions
+    // See: https://github.com/nextstrain/nextclade/issues/707
+    const auto privateSubstitutions = eraseIntersection(query.privateNucMutations.privateSubstitutions,
+      query.privateNucMutations.reversionsOfDeletions);
+
     // TODO: should we also account for result. privateDeletions here
```
corneliusroemer, line 95: Are private deletions counted towards the SNP cluster rule or not, at the moment? Comment sounds like they're not. If they were, it would have to be as ranges, since otherwise a simple 6nt deletion would make the rule fire.

packages/nextclade/src/qc/ruleSnpClusters.cpp
```diff
@@ -72,8 +87,13 @@ namespace Nextclade {
...
-    const auto snpClusters = findSnpClusters(query.privateNucMutations.privateSubstitutions, config);
+    const auto snpClusters = findSnpClusters(privateSubstitutions, config);
```
corneliusroemer, line 96: Because the changes touch only the `ruleSnpCluster.cpp` file, this should have no impact on anything other than that QC rule within Nextclade, correct?

corneliusroemer (approved):
Tested and works as expected, thanks!

packages/nextclade/src/qc/ruleSnpClusters.cpp
ivan-aksamentov: No, deletions are not counted (and were not counted before)

packages/nextclade/src/qc/ruleSnpClusters.cpp
ivan-aksamentov: Right

Issue #712, opened 2022-01-28
BUG[Visual]: Mutations at beginning and end of mut view are invisible due to insufficient padding


corneliusroemer:
When there are mutations in the first or last 0.1% of the genome (e.g. pos 10 in SC2), they are not as visible as they should be because they crash against the edges of the mutation view.

Padding should be increased to make such mutations properly visible.

Consider for example sequence OU187687 in the example dataset, it has lots of mutations in the range 4-30, but these are hardly visible.
![image](https://user-images.githubusercontent.com/25161793/151594299-dafb5b91-1060-40b0-b16d-5f4c38a561d1.png)

![image](https://user-images.githubusercontent.com/25161793/151594305-9da0e53d-0d16-4207-afc2-01b81b11f3c6.png)

ivan-aksamentov:
I tried to improve it a little in https://github.com/nextstrain/nextclade/pull/713, so that at least one could distinguish single mutations at the positions near 1. 

I am not sure more padding would help. Padding will only squeeze the available space making the mutations overlap more.

However in the particular case on the screenshot, the mutations are packed tightly, and overlap, and also obscured by the gray marker of Ns, which makes them even harder to read.

I cannot think immediately about what can be done further here, but glad to hear ideas.

The #295 is probably the ultimate solution here.

PR #713, opened 2022-01-31
fix: shift sequence views such that centered markers are visible

ivan-aksamentov:
Addresses some of the concerns for https://github.com/nextstrain/nextclade/issues/712

Here I shifted sequence (peptide) view's SVG viewBox 1/2 of nucleotide (aminoacid) width to the right. This allows for the markers close to the begging or end of the view to be visible.

This is required after adjustments made for centering markers in 
https://github.com/nextstrain/nextclade/pull/638 https://github.com/nextstrain/nextclade/pull/644 https://github.com/nextstrain/nextclade/pull/709


rneher (approved):

corneliusroemer (approved):
Looks good

PR #722, opened 2022-02-07
fix: exclude reversions of dels when calculating divergence

ivan-aksamentov:
Followup of #710 

We exclude reversions of deletions from mutations considered when calculating divergence. By convention used in the Nextstrain project.


ivan-aksamentov:
Eeek! The first commit did not actually exclude anything. Fixed in the followup commit.

rneher (approved):

rneher:
looks good.

PR #723, opened 2022-02-07
fix: exclude reversions of deletions from private substitutions

ivan-aksamentov:
This excludes reversions of deletions from the array of private substitutions.
After some discussions and trial and error the team decided that they don't belong there.

Previously:

 - divergence in Nextstrain traditionally does not account deletions in any form, so we already excluded reversions of deletions in https://github.com/nextstrain/nextclade/pull/722
 - we've excluded them earlier in SNP clusters QC rule too https://github.com/nextstrain/nextclade/pull/710
 - the only place they show up is the auspice tree
 - they are accounted in totalPrivateMutations

In this PR:
 - [x] removed rev dels globally
 - [x] added rev dels back before attaching mutations to the tree
 - [x] reverted https://github.com/nextstrain/nextclade/pull/710 and https://github.com/nextstrain/nextclade/pull/722, because the filter there is a no-op, since the rev dels are excluded already

The consequences:
 - snp clusters stay the same as it is now (after [#710](https://github.com/nextstrain/nextclade/pull/710) / 1.10.2)
 - divergence decreases, but same as in [#722](https://github.com/nextstrain/nextclade/pull/722)
 - tree still has rev dels
 - output files lose rev dels from `privateNucMutations.privateSubstitutions`, and `privateNucMutations.totalPrivateSubstitutions` decreases

### DIFF
--- a/packages/nextclade/include/nextclade/nextclade.h
+++ b/packages/nextclade/include/nextclade/nextclade.h
@@ -374,16 +374,6 @@ namespace Nextclade {
     return deletion < labeled.deletion;
   }
 
-  // A naive way to erase mutations from a vector of mutations, that are present in another vector of mutations
-  inline safe_vector<NucleotideSubstitutionSimple> eraseIntersection(
-    const safe_vector<NucleotideSubstitutionSimple>& big, const safe_vector<NucleotideSubstitutionSimple>& small) {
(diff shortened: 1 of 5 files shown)


corneliusroemer (approved):
Tested the preview app, had a look at the csv and it looks good.

PR #725, opened 2022-02-08
fix: don't add "truncated" suffix to inserted fragment

ivan-aksamentov:
This is a leftover from the time when the fragment was displayed as a plain string. 

image

At some point the fragment has become a component of its own and the "truncated" text was displayed separately, but I forgot to remove the suffix.

This PR fixes it by removing the concatenation of the suffix to the fragment.


corneliusroemer (approved):
Fixed
image

PR #741, opened 2022-02-24
feat: extend tree extensions node clade attrs to be more descriptive

ivan-aksamentov:
Extends the Nextclade's Augur tree extension object to provide more data (friendly name of the column and description for the tooltip).

Replaces the `meta.extensions.nextclade.clade_node_attr_keys`array of strings with `meta.extensions.nextclade.clade_node_attrs` array of objects in the auspice tree json.

The expected format is as follows:

```json
{
  "meta": {
    "extensions": {
      "nextclade": {
        "clade_node_attrs": [
          {
            "name": "pango_lineage",
            "displayName": "Pango lineage (Nextclade)",
            "description": "Pango lineage as inferred by Nextclade from the nearest neighbour in the reference tree. 98% accurate for recent sequences, for higher accuracy use dedicated pangolin software in UShER or pangoLEARN mode"
            
          },
          {
            "name": "GISAID_clade",
            "displayName": "GISAID clade",
            "description": "Clades as defined by GISAID"
          }
        ]
      }
    }
  }
}
```

All keys are optional, but if at least one key is missing, the object will be ignored. Same as if it's not an object. The entire extension object is optional as well. More fields can be added, if necessary.

Testing: paste the example from above into the existing tree `meta` object and drop it into the web's advanced mode or into cli's `--input-tree`.

### DIFF
--- a/packages/nextclade/include/nextclade/nextclade.h
+++ b/packages/nextclade/include/nextclade/nextclade.h
@@ -636,13 +636,18 @@ namespace Nextclade {
     std::map<std::string, std::string> customNodeAttributes;
   };
 
+  struct CladeNodeAttr {
+    std::string name;
+    std::string displayName;
+    std::string description;
+  };
 
   struct AnalysisResults {
     std::string schemaVersion;
     std::string nextcladeVersion;
     std::uint64_t timestamp;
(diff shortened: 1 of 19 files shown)


corneliusroemer:
Works great, will merge so I can use this on master with a dataset that enables this mode.

Issue #745, opened 2022-03-03
BUG: Mutation at position 1 reported as at position 0


corneliusroemer:
Trying out synthetic lineage sequences, I was startled to find that one of them got a nucleotide mutation reported at position 0, which shouldn't be possible in the front end, since there we use 1-indexing.

<img width="333" alt="image" src="https://user-images.githubusercontent.com/25161793/156611323-aee71e05-20d8-45c0-8c60-143722f4f9b3.png">

To reproduce, try this sequence:
https://pastebin.com/raw/6DBJuxiz

It seems like only the 1st nuc is treated wrong. Since when I change the first two nucs to C, the first is reported as `0C`, second as `2C`

Some coordinate system mishap, I think. Worth checking what happens if there's a mutation at the last position. (I just did and it looks good for now).

corneliusroemer:
It could be that the problem is restricted to the front end/web app, since Auspice shows correct mutations:
<img width="580" alt="image" src="https://user-images.githubusercontent.com/25161793/156614059-b5ec94cc-c5c4-434a-978b-3601626e2425.png">

ivan-aksamentov:
@corneliusroemer This should be fixed in https://github.com/nextstrain/nextclade/pull/747. Please double check.

PR #747, opened 2022-03-07
fix(web): ensure correct rendering of mutation at position 0

ivan-aksamentov:
Resolves https://github.com/nextstrain/nextclade/issues/745

There was a redundant null check in the NucleotideMutationBadge component that was causing position of a mutation at first nuc to be ignored (the number 0 is considered nullish in js). This check is redundant anyway, because the position is never null, so I removed it. AminoacidMutationBadge component does not have this check, so it should behave correctly.


corneliusroemer:
Beautiful, it works!
image

image

corneliusroemer (approved):

PR #763, opened 2022-04-05
feat: discard bad seed positions as soon as possible

corneliusroemer:
If my reasoning is correct, there's an off-by-one error in the original code. We could abort already one mismatch earlier. Hence I replaced `<` with `<=`.

Would be good if @rneher could check the logic. I'm pretty sure this is correct.

I've added unit tests, it all works as expected.

This should speed up seed matching by around 20%, as with 3 mismatches allowed, the expected number of checks before aborting was up to now 5, and now it's 4.

### DIFF
--- a/packages_rs/nextclade/src/align/seed_match.rs
+++ b/packages_rs/nextclade/src/align/seed_match.rs
@@ -22,8 +22,8 @@ pub fn seed_match(kmer: &[Nuc], ref_seq: &[Nuc], start_pos: usize, mismatches_al
         tmp_score += 1;
       }
 
-      // TODO: this speeds up seed-matching by disregarding bad seeds.
-      if tmp_score + mismatches_allowed < pos {
+      // this speeds up seed-matching by disregarding bad seeds.
+      if tmp_score + mismatches_allowed <= pos {
         break;
       }
     }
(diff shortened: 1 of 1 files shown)


rneher:
yes, there was an off-by-one error (as `tmp_score` is incremented before `pos`). But we should then also change the default `mismatches_allowed` to preserve the old behavior. To speed things up, we could break out of the loop once `max_score>kmer_len - mismatches_allowed`.

packages_rs/nextclade/src/align/seed_match.rs
```diff
@@ -43,3 +47,94 @@ pub fn seed_match(kmer: &[Nuc], ref_seq: &[Nuc], start_pos: usize, mismatches_al
...
 }
+
+#[cfg(test)]
+mod tests {
+  use super::*;
+  use crate::io::nuc::to_nuc_seq;
+  use pretty_assertions::assert_eq;
+  use rstest::rstest;
+
+  #[rstest]
```
corneliusroemer, line 58: Tests here could still be useful

corneliusroemer:
This is closed [in #stripe](https://github.com/nextstrain/nextclade/pull/795)

PR #777, opened 2022-05-05
fix(web): display codon 0 correctly in the peptide context

ivan-aksamentov:
Codon 0 (1 in one-based indexing) was not displayed correctly in the peptide context due to `0` treated as `false` in the conditional statement.


corneliusroemer:
Yup, fixed!

PR #780, opened 2022-05-09
fix: strip gap characters from qry seq when reading in

corneliusroemer:
We need to strip `-` characters from the query sequence as some of our algorithms assume the query sequence doesn't contain these.

I created a new function to ungap straight when reading in, maybe there's a better way to do this. We may also ungap the reference sequence, just in case. Or at least warn if it contains gaps, since that would be a problem for our algorithms, too.


ivan-aksamentov:
Superseded by #781

PR #786, opened 2022-05-17
fix: prepend new node in Auspice tree

corneliusroemer:
New node is attached at bottom rather than top in line with Auspice/Augur convention. Inserting at front of vec is not great for performance, but this is probably not limiting anyways. Otherwise, maybe we can put children from back to front in array rather than front to back.

Screenshot 2022-05-17 at 15 28 45


ivan-aksamentov:
@rneher Could you please double-check this. I remember there was particular preferred order of attaching child nodes back in the day. Is the prepending proposed here better than appending or maybe some other ordering is needed entirely?

Issue #653, opened 2021-12-15
Duplicate clade label shown in auspice tree, when clade consists of 1 tip only


corneliusroemer:
When a nextclade reference tree has a clade that is defined by only 1 single tip (like 21L in the current version), if any new nodes are attached to that node, the clade label seems to be duplicated causing a visual bug in the Auspice view [note the duplicated 21L]: 
![image](https://user-images.githubusercontent.com/25161793/146267613-fea8f2eb-869e-4e6a-97e2-1babb42ac690.png)

corneliusroemer:
Fixed in v2

PR #787, opened 2022-05-17
fix: ref leaf was missing in tree, fixes label duplication during node attachement

corneliusroemer:
fixes https://github.com/nextstrain/nextclade/issues/653
1. Reference leaf was missing when new node was attached to leaf.
2. Fixes label duplication when new node is attached (old known problem)

Tested locally, seems to work.

image

### DIFF
--- a/packages_rs/nextclade/src/tree/tree_attach_new_nodes.rs
+++ b/packages_rs/nextclade/src/tree/tree_attach_new_nodes.rs
@@ -49,7 +49,6 @@ fn attach_new_node(node: &mut AuspiceTreeNode, result: &NextcladeOutputs) {
 
 fn add_aux_node(node: &mut AuspiceTreeNode) {
   debug_assert!(node.is_ref_node());
-  println!("Adding aux node to {}", node.tmp.id);
 
   let mut aux_node = node.clone();
   aux_node.branch_attrs.mutations.clear();


packages_rs/nextclade/src/tree/tree_attach_new_nodes.rs
```diff
@@ -52,6 +52,9 @@ fn add_aux_node(node: &mut AuspiceTreeNode) {
 
   let mut aux_node = node.clone();
   aux_node.branch_attrs.mutations.clear();
+  // Remove other branch attrs like labels to prevent duplication
+  aux_node.branch_attrs.other = serde_json::Value::default();
```
ivan-aksamentov, line 56: This removes all attributes that are not explicitly listed in the [`TreeBranchAttrs` struct](https://github.com/nextstrain/nextclade/blob/a171903b93999dca5ef12ef01ff1a869b924ae31/packages_rs/nextclade/src/tree/tree.rs#L37-L43). What if users would like to keep some of the fields they added? Should we instead remove only a specific field or fields?

It can be done as follows:

 - Add optional field or fields to the struct, and mark them "skip if `None`":

   ```rust
   #[serde(skip_serializing_if = "Option::is_none")]
   pub field_name: Option<String>
   ```

 - Set these fields to `None` when needed. They will be then skipped during deserialization.

corneliusroemer: Good question.

The tip is on a zero length branch, so all branch attrs I can think of should be on the actual branch leading to the new internal node.

That branch has moved so should all the labels, I'd say. Can't think of any labels we use that should be on both sides.

I'd say if the need arises we can still do the implementation you suggest above, but right now it'd not add anything.

PR #793, opened 2022-05-19
chore: upgrade rust deps

ivan-aksamentov:


ivan-aksamentov:
We'll do this later when things calm down a bit after release

PR #795, opened 2022-05-19
Banded alignment with flexible bandwidth

rneher:
Our previous alignment used one fixed bandwidth for the entire length of the banded alignment. 

this PR implements (mostly work by @corneliusroemer ) a banded alignment where the bandwidth and the relative shift of reference and query along the reference sequence. This makes the alignment much faster and much more memory efficient if there are large insertions or deletions in the sequence. 

The alignment band is determined by seed matches. A group of 4 seeds `s1 -- s2 -- s3 -- s4` determine the shift and bandwidth between seeds `s2` and `s3` by using the max and min qry/ref shift at these four seeds as band. The band is further broadened by `excess_bandwidth`.

Compared the previous version, the seed matching is also more robust by restricting the reference positions that are being searched and enforcing monotonicity of  seed matches. 

Prior to merge, we need
  * [x]  create full alignment matrix for short sequences, clean up/remove corresponding legacy code
  * [x]  reduce the maximal capacity for the aligned sequences
  * [x]  add toggle for left/right align, add corresponding CLI flag pair.

Possible other improvements include 

 * replace score matrix by two vectors, return optimal score from score_matrix
 * allow 1-d access to path matrix.


packages_rs/nextclade/src/align/seed_alignment.rs
```diff
@@ -208,6 +210,10 @@ pub fn create_stripes(
   for slice in shifts.windows(4) {
     let min = slice.iter().min().unwrap();
     let max = slice.iter().max().unwrap();
+    let width = max - min;
+    if width as usize> max_indel {
```
corneliusroemer, line 211: Max indel is now not just a single indel but the total shift over 3*seed spacing. So we may want to increase max_indel by default.

PR #797, opened 2022-05-20
feat(cli): take all genes if --genes is not provided

ivan-aksamentov:
This modifies behavior of various combinations of `--genemap` and `--genes` flags and makes it consistent in Nextclade CLI and Nextalign CLI.

The new behavior is as follows:

| --genemap  | --genes |                 behavior                   |
|------------|---------|--------------------------------------------|
|     +      |    +    | Take only specified genes                  |
|     +      |         | Take all genes                             |
|            |    +    | Error                                      |
|            |         | Skip translation and codon penalties       |

Notably, if `--genemap` is provided, but `--genes` is not, then all genes will be taken into consideration.


rneher:
looks good to me!

PR #800, opened 2022-05-20
feat: add gap alignment side enum param

ivan-aksamentov:
For mutually exclusive options.

The CLI args are:

```
--gap-alignment-side=left
--gap-alignment-side=right
```

Perhaps better to call is "gap placement side" or something like this?


ivan-aksamentov:
TODO:
 - [ ] remove the old param pair
 - [ ] make sure that the default is set correctly
 - [ ] make sure that the usage of the param in the algorithm is correct

PR #801, opened 2022-05-20
feat: dynamic alignment params

ivan-aksamentov:
NOTE: this PR targets branch [experiment-stripes-updated](https://github.com/nextstrain/nextclade/tree/experiment-stripes-updated).

This PR allows both CLI and Web to use the new optional field `alignmentParams` in `virus_properties.json`. These are equivalent to the params provided through command line flags. All fields are optional.

The mechanism of deducing final alignment params is as follows:

 - web:
    - the default params object is created
    - `alignmentParams` is read from `virus_properties.json` and merged into the defaults - every defined field overrides the corresponding field

 - CLI: 
    - the default params object is created
    - `alignmentParams` object is read from `virus_properties.json`, merged into the default, and every defined field overrides the default
    - params from CLI are read, merged into what we've got so far - every passed CLI param overrides the corresponding field

Caveats:
 - with all the struct magic involved, I could not figure out how to keep the default values printed in CLI help, so these are gone. I might be able to find a way of how to keep them later.

An example snippet to paste into the root object of the `virus_properties.json`:

```json
  "alignmentParams": {
    "max_indel": 5000,
    "seed_spacing": 500
  },
```


corneliusroemer:
Will the schema version need to be bumped? I have it at 1.10.0 at the moment almost everywhere I think.

corneliusroemer:
Works well, will merge

PR #802, opened 2022-05-20
fix(web): display codon 0 correctly in peptide context

ivan-aksamentov:
This is an extended repeat of https://github.com/nextstrain/nextclade/pull/777, which was accidentally wiped out sometimes during merges.


rneher:
looks like it works!

PR #805, opened 2022-05-21
fix: make GapAlignmentSide values in JSON to resemble the ones in CLI

ivan-aksamentov:
Uses [serde's `rename_all` container attribute](https://serde.rs/container-attrs.html#rename_all) to convert variants of `enum GapAlignmentSide` to kebab case during (de)serialization, e.g. in `virus_proprieties.json`.

The enum variants in Rust are by convention in PascalCase, and serde uses them as is. But we want to make the values in `virus_proerties.json` to resemble the values in CLI arguments, and clap seems to be converting enums to kebab case.

If merged, this change requires simultaneous adjustment of datasets using the `gap_alignment_side` param.


ivan-aksamentov:
TODOs:
 - [ ] simultaneously adjust `gap_alignment_side` param in `virus_proprieties.json` in datasets

PR #806, opened 2022-05-21
Better genemap warnings

corneliusroemer:
Adds two warnings to alert user if something undesired happens:
- warn if one or more --genes requested are not actually found in the parsed --genemap
- warn if non-empty genemap is supplied (commented lines are not counted) but no genes are found


ivan-aksamentov:
I think these should be errors, because:
 - you never mean to request a non-existent gene - it's always a typo or a bug in a pipeline on user's side
 - it does not make sense to have gene map with rows that don't contain genes, at least for nextclade, so that's likely a bug in userland as well

When you first encountered this, I believe that you'd be able to resolve problem faster if it would just refuse to accept the broken gene map.

ivan-aksamentov:
In terms of code, you could avoid allocating a vector of fauly genes and could just `.join(", ")` them. The printing could be simplified as well: `format!("blabla {faulyGenes}`")`.

ivan-aksamentov (approved):

packages_rs/nextclade/src/io/gff3.rs
```diff
@@ -53,10 +54,21 @@ fn read_gff3_file_impl<P: AsRef<Path>>(filename: &P) -> Result<GeneMap, Report>
...
+  if let Ok(genemap) = &genemap_result {
+    if genemap.len() == 0 && records.len() != 0 {
+      warn!(
+        "No valid gene entries found in genemap with {} records. No genes will be used.",
+        records.len()
+      );
+    }
+  }
+
+  genemap_result
```
ivan-aksamentov, line 71: This snippet is also a bit confusing and verbose.
You typically want to get rid of `Result` and return as soon as possible, to not carry arround the checks, so something like this:

```
let genemap = (...).collect()?; // get rid of Result
if genemap.len() { ... } // no check needed
Ok(genemap)
```

corneliusroemer:
> The second check could also be done when converting individual rows, instead of after the fact. In this case you could even tell which row has no gene, for a more actionable warning/error.

The reason I don't do this is because some lines may contain things that aren't genes. Just a gene map that contains no genes at all makes not much sense so thought it would be worth warning.

Thanks for your very helpful code review comments. I addressed them. Please merge if you're satisfied.

ivan-aksamentov (approved):

PR #808, opened 2022-05-21
Allow custom score weight for frame shift and stop codon QC rule

corneliusroemer:
- feat: add new field "score_weight" to qcconfig, with default 75
- feat: use custom score_weight in fs/stop qc rules

We already have a custom score weight field in qc config for SNP clusters.

I now add these fields also to frame shift and stop rules, so they can be weighted down for monkeypox which has a big genome and thus more chances for both.

If the fields are not included, they default to the old value of 75, so this is backwards compatible with respect to old qc_config.jsons

New qc_configs also work with old software versions. These fields are simply ignored. That's really nice with `serde`!

How does scheming of our config files work? Do I need to bump the version and change the schema somewhere? Maybe we could warn or trace or something like that if some config values are missing?


ivan-aksamentov:
> How does scheming of our config files work? 

The version is incremented when there are breaking changes.

> Do I need to bump the version and change the schema somewhere?

No, because as you said there are not breaking changes.

ivan-aksamentov:
I changed the code that produces default values. It is important to not only please serde or clap with annotations, but to make sure the rest of the code also receives the correct defaults.

I don't quite remember if the defaults are used somewhere else, but that would be my first reflex developed after ling painful years of writing C++ constructors.

corneliusroemer:
Excellent, thanks!

ivan-aksamentov:
> I now add these fields also to frame shift and stop rules, so they can be weighted down for monkeypox which has a big genome and thus more chances for both.

@corneliusroemer @rneher As a more permanent solution here, should I pass ref genome length to QC, so that you could weight the scores (for any rules, not only these two). If a scale relative to genome size is needed, then it would be more sound to the value itself, rather than some made-up weight. Plus it will work automagically for every new virus and require no modifications to the datasets. I cannot tell how much sense it makes scientifically though, it's just my imagination working.

PR #809, opened 2022-05-21
fix: rev strand dir `-` is parsed as `(-)`, now detected correctly

corneliusroemer:


ivan-aksamentov:
I believe that serialization should be fixed instead if possible. Also `gene.strand` should be an `enum` in the first place.

PR #811, opened 2022-05-23
feat(web): configure sequence markers height

ivan-aksamentov:
This adds toggles to the "Settings" dialog, which allow to toggle markers in sequence views between full-height, half-height at the bottom, half-height at the top and turned off entirely. The toggles are separate for missing, gaps, mutations and unsequenced markers.

The values are remembered in local storage and persist across app launches.

Right now only the nuc view is configurable and only for 4 of hese categories of vertical markers (and not frame shifts for example), but we can add similar toggles to peptide view and other markers later.

Known issues:
 - [ ] only the last blue toggle is highlighted with white text. Something weird there.
 - [ ] could not figure out how to make a horizontal form (labels on same line). Reactstrap makes this surprisingly difficult.

### DIFF
--- a/packages_rs/nextclade-web/src/state/seqViewSettings.state.ts
+++ b/packages_rs/nextclade-web/src/state/seqViewSettings.state.ts
@@ -1,0 +1,77 @@
+import { atom } from 'recoil'
+import { ErrorInternal } from 'src/helpers/ErrorInternal'
+
+import { persistAtom } from 'src/state/persist/localStorage'
+
+export enum SeqMarkerHeightState {
+  Off = 'Off',
+  Top = 'Top',
+  Bottom = 'Bottom',
+  Full = 'Full',
+}
+
+export const SEQ_MARKER_HEIGHT_STATES = Object.keys(SeqMarkerHeightState)
+
+export function seqMarkerHeightStateToString(val: SeqMarkerHeightState) {
+  return val.toString()
+}
+
+export function seqMarkerHeightStateFromString(key: string) {
+  // prettier-ignore
+  switch (key) {
+    case 'Top': return SeqMarkerHeightState.Top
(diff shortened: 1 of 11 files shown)


packages_rs/nextclade-web/src/state/seqViewSettings.state.ts
```diff
@@ -0,0 +1,77 @@
...
+    case SeqMarkerHeightState.Off:
+      return { y: 0, height: 0 }
+  }
+  throw new ErrorInternal(`getSeqMarkerDims: Unknown 'SeqMarkerHeightState' variant: '${state}'`) // eslint-disable-line @typescript-eslint/restrict-template-expressions
+}
+
+export const seqMarkerMissingHeightStateAtom = atom<SeqMarkerHeightState>({
+  key: 'seqMarkerMissingHeight',
+  default: SeqMarkerHeightState.Top,
+  effects: [persistAtom],
```
ivan-aksamentov, line 47: The defaults can be configured here

packages_rs/nextclade-web/src/state/seqViewSettings.state.ts
```diff
@@ -0,0 +1,77 @@
...
+      return { y: -10, height: 10 }
+    case SeqMarkerHeightState.Bottom:
+      return { y: 10, height: 10 }
+    case SeqMarkerHeightState.Full:
+      return { y: -10, height: 30 }
+    case SeqMarkerHeightState.Off:
+      return { y: 0, height: 0 }
+  }
+  throw new ErrorInternal(`getSeqMarkerDims: Unknown 'SeqMarkerHeightState' variant: '${state}'`) // eslint-disable-line @typescript-eslint/restrict-template-expressions
+}
```
ivan-aksamentov, line 42: Sizes can be tweaked here

PR #814, opened 2022-05-24
fix(web): treat dataset default gene field as optional

ivan-aksamentov:
When `defaultGene` was not set in the dataset, causing "Gene is missing in gene map" text appearing in the sequence view, due to gene view being switched to peptide view for `undefined` gene.

Let's treat the `defaultGene` field as optional and, if it's undefined, default to nuc view.


corneliusroemer (approved):
It works great, let's merge, this is an important bug fix.

PR #815, opened 2022-05-25
Support reverse (minus) strands

ivan-aksamentov:
This PR adds changes that allow for supporting of reverse (negative) strands in viral genomes.


rneher:
Thanks, Ivan. This seems to work as expected. The only remaining issue now is that the nucleotide numbering should go in reverse and a hint for the user that this is the reverse complement:

image

rneher:
the nucleotide numbering still needs adjusting:
image

note that the mutated base is 54126 and the numbering needs to decrease from left to right (reverse strand...)

ivan-aksamentov:
@rneher Yes, that's in the works, along with some refactoring

rneher:
I am getting null-pointer exception in the preview app.

corneliusroemer:
Still getting that null pointer error
image

ivan-aksamentov:
The real error is in the browser console. The null pointer error here is an emulation of Rust panics (web platform does not have a concept of panics).

In this case the panic is caused by an array overrun due to mistakes in calculating indices. It's not very useful to run the web app before the algorithm is fixed.

rneher:
I added tests and fixed a couple of errors.

rneher:
I'll dig into the actual frame shift code later.

PR #818, opened 2022-05-25
fix(web): ensure dataset is read from URL

ivan-aksamentov:
This ensures that the dataset is properly changed if the launch happens from URL.

With this, the previous dataset, persisted in local storage is used, which causes analysis errors if the requested dataset is not the same as the previously used dataset.


rneher:
LGTM

PR #820, opened 2022-05-25
fix(web): ensure seq name column and tooltip are shown for failed seq

ivan-aksamentov:
For sequences that fail alignment the name column was blank and no tooltip was shown. This was due to incorrect conditional logic in the component.

Now the name column and its tooltip are displayed correctly.


rneher:
LGTM

PR #826, opened 2022-05-30
fix: ensure aa muts are attached to the tree

ivan-aksamentov:
This fixes an omission: only nuc mutations were attached to the tree JSON object, and aa mutations were missing.

### DIFF
--- a/packages_rs/nextclade/src/analyze/aa_sub.rs
+++ b/packages_rs/nextclade/src/analyze/aa_sub.rs
@@ -1,4 +1,4 @@
-use crate::io::aa::Aa;
+use crate::io::aa::{from_aa, Aa};
 use crate::io::letter::Letter;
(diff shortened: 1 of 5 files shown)


rneher:
there is something funky going on. In the SARS-CoV-2 example data in the 21L clade, this branch has 29 amino acid mutations mapped to it. It shouldn't. 

image


and this branch should have two mutations in ORF1a (488 and 556) that are not shown as mutations on the branch:

image

corneliusroemer:
I can confirm the bug, I'll quickly review code to see if I can find the issue :)

corneliusroemer:
Fix works! I'll merge to get this bug out of the way. If there's another problem we can do another PR. If Ivan wants to refactor, he can do so independently ;) We wrote this code yesterday mostly together anyways.
image

PR #827, opened 2022-05-30
feat(web): display nuc and aa insertions in sequence views

### LINKED ISSUE #769: ENH: Surface insertions in mutation view
Apparently there's a Danish BA.2 lineage with an insertion around S:212. BA.1 had an insertion, so did other lineages.

It would be nice if we could show them graphically in the alignment/mutation view, not just in the table.

We'd want to make it visually clear that these are insertions - but I'm sure we can figure something out.

It would be a great addition to see insertions in detail view the same way we show deletions in detail view in tooltips.

I can think of a few ways to differentiate insertion mutation strips from normal mutations: we could make them half height, or have a little bar through them, like we do with frameshifts.

ivan-aksamentov:
This adds display of insertions in nucleotide and aminoacid sequence views

Resolves #769


rneher (approved):

corneliusroemer:
I'd change colour to something less "errory" otherwise perfect!

PR #828, opened 2022-05-30
feat(web): display total number of markers and limit in the message

ivan-aksamentov:
This adds some more information to the message appearing instead of sequence view when the limit of markers is exceeded. Now the message shows how many markers it is in total to draw and what is the current limit.

This gives users some more actionable data on how to adjust the limit.


rneher (approved):
Why don't you increase the default to 1000 and merge?

ivan-aksamentov:
@rneher The second sequence in the sample data will still not be rendered. It has 2146 markers (plus some more when insertions https://github.com/nextstrain/nextclade/pull/827 are merged). Is that alright?

PR #857, opened 2022-06-14
feat(cli): make input fasta arg positional, allow multiple files

ivan-aksamentov:
- [x] Removes `--input-fasta` flag in favor of providing input sequence file names as positional arguments
 - [x] Multiple files can be provided by passing multiple positional arguments
 - [x] Files can be plain fasta or compressed
 - [x] Different files can have different compression
 - [x] Read from stdin if no positional arguments provided
 - [x] Warn if no positional arguments provided and stdin is a TTY
 - [x] Output to stdout if filename is `-`

One quirk is that now if `--output-basename` is not provided, and input has multiple fasta files it's unclear which one to use for finding basename of output files. I opted for:
 - taking the name of the file if only one provided, just like before
 - hardcoded string "nextclade"/"nextalign" if there are multiple input files or no files (stdin)

Examples:

```
# Single file
nextclade run -D dataset/ -O out/ 1.fasta

# Multiple files
nextclade run -D dataset/ -O out/ 1.fasta 2.fasta 3.fasta

# Multiple files with different compression
nextclade run -D dataset/ -O out/ 1.fasta 2.fasta.gz 3.fasta.zst

# Wildcard (expanded by shell)
nextclade run -D dataset/ -O out/ *.fasta

# Wildcard (expanded by shell), including compressed files
nextclade run -D dataset/ -O out/ *.fasta*

# Reading from standard input (only plain, uncompressed fasta)
cat 1.fasta 2.fasta | nextclade run -D dataset/ -O out/

# Reading from standard input, compressed (external decompression)
xzcat 1.fasta.xz 2.fasta.xz | nextclade run -D dataset/ -O out/

# Output to stdout if filename is "-"
nextclade run -D dataset/ --output-tsv=- 1.fasta | process_nextclade_tsv_further

# 3-stage Pipeline
xzcat *.fasta.xz | 
nextclade run -D dataset/ --output-tsv=- 1.fasta | 
process_nextclade_tsv_further

# Nextalign works as well
xzcat *.fasta.xz | 
nextalign run -r ref.fasta -m genemap.gff -o - |
process_aligned_fasta
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1373,6 +1373,7 @@ name = "nextclade"
 version = "2.0.0-alpha.12"
 dependencies = [
  "assert2",
+ "atty",
  "auto_ops",
  "bio",
  "bio-types",
--- a/packages_rs/nextclade-cli/src/cli/common.rs
+++ b/packages_rs/nextclade-cli/src/cli/common.rs
@@ -1,0 +1,17 @@
+use nextclade::io::fs::{basename_maybe, extension};
+use std::path::PathBuf;
+
+pub fn get_fasta_basename(input_fasta: &[PathBuf]) -> Option<String> {
+  match input_fasta {
+    [single_fasta] => {
+      let base_name = basename_maybe(single_fasta)?;
+      if extension(&base_name).map(|ext| ext.to_lowercase()) == Some("fasta".to_owned()) {
+        // Additionally handle cases like `.fasta.gz`
+        basename_maybe(&base_name)
+      } else {
+        Some(base_name)
+      }
+    }
+    _ => None,
+  }
+}
--- a/packages_rs/nextclade-cli/src/cli/mod.rs
+++ b/packages_rs/nextclade-cli/src/cli/mod.rs
@@ -1,3 +1,4 @@
+pub mod common;
 pub mod nextalign_cli;
 pub mod nextalign_loop;
 pub mod nextalign_ordered_writer;
--- a/packages_rs/nextclade-cli/src/cli/nextalign_cli.rs
+++ b/packages_rs/nextclade-cli/src/cli/nextalign_cli.rs
@@ -1,4 +1,5 @@
-use clap::{AppSettings, ArgEnum, ArgGroup, CommandFactory, Parser, Subcommand, ValueHint};
+use crate::cli::common::get_fasta_basename;
+use clap::{AppSettings, ArgEnum, CommandFactory, Parser, Subcommand, ValueHint};
 use clap_complete::{generate, Generator, Shell};
 use clap_complete_fig::Fig;
 use clap_verbosity_flag::{Verbosity, WarnLevel};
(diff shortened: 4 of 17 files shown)


tsibley:
_Responding to @ivan-aksamentov's [request for feedback in Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1655200598022889)._

Overall this looks like a more standard set of behaviours for command-line programs, so :+1: from me.

Re: output to stdout: the `--output-tsv=/dev/stdout` and `-o /dev/stdout` syntax will only work on systems which support those pseudo-devices. Linux, for example, symlinks those to `/proc/self/fd/1` and `…/fd/2`, but the proc fs isn't available on other systems, though there may be equivalents. It seems worth considering how stdout is specified for those options on macOS and Windows. (On OSes which don't support the devices, Bash emulates them but only when they appear in redirections, not in arguments, see excerpt from the man page below.)


<details><summary><i>Bash special filenames in redirections</i></summary>

```
Bash handles several filenames specially when they are used in redirec‐
tions, as described in the following table.  If the operating system on
which bash is running provides these special files, bash will use them;
otherwise it will emulate them internally with the  behavior  described
below.

      /dev/fd/fd
             If  fd  is  a valid integer, file descriptor fd is dupli‐
             cated.
      /dev/stdin
             File descriptor 0 is duplicated.
      /dev/stdout
             File descriptor 1 is duplicated.
      /dev/stderr
             File descriptor 2 is duplicated.
      …
```

</details>

On Slack you wrote:

> I see only one drawback related to reading from stdin by default: if you forget to add files, you'd be staring at a blank screen and/or clean cluster logs for a little while. Happens to me all the time…

Agreed that this is a common tripping hazard. The traditional way to handle it is to check if stdin is a TTY when input is stdin. If it is, then error (if there's no chance that anyone wants to type directly into `nextclade`/`nextalign`) or print a loud warning to stderr (if there is a chance).

ivan-aksamentov:
@tsibley Thanks! That's very valuable comments.

> Re: output to stdout

What special handling I could add to make it cross platform? I know that some well-known cross-platform tools (e.g. curl) can accept `-` as a file name and write to stdout in this case. Will `if "-", then printf(), else fopen()` kind of logic be enough here?

Looks like curl does a little bit of magic regarding ascii/binary output and checking whether user requested stdout explcitly, to avoid "binary gunk":
https://github.com/curl/curl/blob/4572489c5946b5a99d20d6b3692282b07fd4eccd/src/tool_operate.c#L1208-L1218
All our outputs are UTF-8 and writing to stdout is requested explicitly, so I guess it should be fine to just check if the filename is "-" and copy the strings to buffered `std::io::stdout`.


> check if stdin is a TTY when input is stdin. If it is, then error

Nice! I might try it. For example with [atty](https://github.com/softprops/atty). I hope there will be no problems with docker and its various `-it` options.

tsibley:
Yep, using `-` as a special value to mean "stdout" is good and very conventional. Agreed you can just spit out UTF-8 to stdout.

rneher (approved):

Issue #865, opened 2022-06-16
NextClade 2.0 on Safari: "An error has occurred: ReferenceError Can't find variable: Worker"


sacundim:
I just saw the Nextclade 2.0 announcement, downloaded some SARS-CoV-2 sequences off GenBank, tried to load them into the web interface, and got the error below right away.

Environment:

* Safari Version 15.5 (17613.2.7.1.8) 
* MacOS Monterey Version 12.4
* Intel Mac

```
An error has occurred: ReferenceError
Can't find variable: Worker

@https://clades.nextstrain.org/_next/static/chunks/launcherWebWorker.eed3e2c818a5e20b.js:1:9275
u@https://clades.nextstrain.org/_next/static/chunks/444.ca2625476ee8666b.js:1:24019
@https://clades.nextstrain.org/_next/static/chunks/444.ca2625476ee8666b.js:1:25306
J@https://clades.nextstrain.org/_next/static/chunks/launcherWebWorker.eed3e2c818a5e20b.js:1:8684
d@https://clades.nextstrain.org/_next/static/chunks/launcherWebWorker.eed3e2c818a5e20b.js:1:8882
@https://clades.nextstrain.org/_next/static/chunks/launcherWebWorker.eed3e2c818a5e20b.js:1:8941
Promise@[native code]
@https://clades.nextstrain.org/_next/static/chunks/launcherWebWorker.eed3e2c818a5e20b.js:1:8833
u@https://clades.nextstrain.org/_next/static/chunks/444.ca2625476ee8666b.js:1:24019
@https://clades.nextstrain.org/_next/static/chunks/444.ca2625476ee8666b.js:1:25306
J@https://clades.nextstrain.org/_next/static/chunks/launcherWebWorker.eed3e2c818a5e20b.js:1:8684
d@https://clades.nextstrain.org/_next/static/chunks/launcherWebWorker.eed3e2c818a5e20b.js:1:8882
@https://clades.nextstrain.org/_next/static/chunks/launcherWebWorker.eed3e2c818a5e20b.js:1:8941
Promise@[native code]
@https://clades.nextstrain.org/_next/static/chunks/launcherWebWorker.eed3e2c818a5e20b.js:1:8833
u@https://clades.nextstrain.org/_next/static/chunks/444.ca2625476ee8666b.js:1:24019
@https://clades.nextstrain.org/_next/static/chunks/444.ca2625476ee8666b.js:1:25306
[...]

Issue #779, opened 2022-05-09
BUG: Out of bounds memory access (evaluating 'd.get(t)(e,r,n)') on M1 Safari only (not on Chrome)


elginakin:
Hello. When attempting to run an Influenza B assembly and workflow through nextclade, my job fails continuously and I am met with the following message ``` Out of bounds memory access (evaluating 'd.get(t)(e,r,n)') ```. Moreover, this error occurs with any example workflow I have tried to run. Not sure what to make of it. Any help would be appreciated. Thanks. 

![Screen Shot 2022-05-09 at 8 30 48 AM](https://user-images.githubusercontent.com/63241361/167410680-ac594fa3-f28e-457e-8e3f-cfe5e450e6ca.png)

corneliusroemer:
Just reproduced this on Chrome and Safari on an iPhone 13, iOS 15.4.1

Noteworthily, on Chrome as well, not just Safari.

But sure, mobile isn't our focus.

ivan-aksamentov:
All browsers on iOS, including Chrome, use WebKit engine, because Apple requires so. So they all are just Safari with different skins.

corneliusroemer:
Aha, that explains it!

ivan-aksamentov:
@eakin1 If time allows, could you please try to reproduce this issue again with the new Nextclade version 2.0?

It looks roughly the same, but the internals are completely rewritten. We haven't released it yet, but here is a testing version:
https://master.clades.nextstrain.org/

We are trying to catch last bugs before the release, so your help is highly appreciated!

ivan-aksamentov:
Wow! New error... That is unexpected. Thank you for checking!

ivan-aksamentov:
Oh no! Looks like Safari does not support nested WebWorkers, which we need for v2.

> Safari unfortunately as of now does not support spawning Web Workers from inside other Web Workers.
https://stackoverflow.com/a/65099653

It's been 7 years and Apple doesn't seem to be particularly interested:
https://bugs.webkit.org/show_bug.cgi?id=25212

ivan-aksamentov:
The original issue is fixed in Nextclade v2 and regarding Safari - we decided to drop supporting it due to lack of required browser technologies in it and lack of technical resources on our side.

I added a warning message to notify users of unsupported browsers in https://github.com/nextstrain/nextclade/pull/866

If someone decides to fix Safari - contributions are welcome.

PR #866, opened 2022-06-16
feat(web): add warning for unsupported browsers

ivan-aksamentov:
Related: https://github.com/nextstrain/nextclade/issues/779, https://github.com/nextstrain/nextclade/issues/865

Adds an alert on top of the page if a browser does not meet these criteria:

https://github.com/nextstrain/nextclade/blob/1ec1c2b0726221b301e4d31e8443d792852f312c/packages_rs/nextclade-web/src/components/Common/BrowserWarning.tsx#L13-L17

Firefox and Chrome versions are chosen semi-randomly among versions released circa 2017 (5 years ago), which should give some room. Chrome here includes all the Chromium-based derivatives which don't modify User-Agent string too much (including Brave for example).

Just to be sure, I added Edge 79 as the first Chromium-based version.

Safari support is dropped entirely due to lack of support of nested webworkers: https://bugs.webkit.org/show_bug.cgi?id=25212

### DIFF
--- a/packages_rs/nextclade-web/package.json
+++ b/packages_rs/nextclade-web/package.json
@@ -87,6 +87,7 @@
     "awesomplete": "1.1.5",
     "axios": "0.27.1",
     "bootstrap": "4.6.1",
+    "bowser": "2.11.0",
     "classnames": "2.3.1",
     "compare-versions": "4.1.3",
     "core-js": "3.22.2",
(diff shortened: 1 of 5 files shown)


corneliusroemer:
Nice! I think we should make the notice much "louder" though - not orange but red, make it bold, maybe not just at the top but as an overlay that needs to be clicked away?

Also, we should probably not even allow the user to proceed with "run", blocking the button.

Right now, one really needs to read that orange text at the top, that's the only warning. Otherwise it errors as before.

image

PR #869, opened 2022-06-17
feat(web): show dataset machine-readable name in dataset selector

ivan-aksamentov:


corneliusroemer:
This is good - can be merged I think

PR #885, opened 2022-06-22
fix: avoid large allocations during alignment

ivan-aksamentov:
This:
 - [x] adds `--min-match-rate` alignment parameter (0.3 by default) which sets the expected rate of seed matching over total number of attempted seeds.


rneher:
I think this is useful and can be merged. We can add attempt to rescue reverse complemented sequences later. I mentioned yesterday that it would be useful to have an additional catch of at least 2 seed matches. This would likely never trigger, but if somebody sets `seed-spacing` to a large number and `min-match-rate` to a small number, you might get situations where you have very few seeds. I wouldn't accept sensible behavior with one seed only.

PR #887, opened 2022-06-23
feat: retry with reverse complement when seed matching fails

ivan-aksamentov:
Adds flag `--retry-reverse-complement` which enables additional attempt of seed matching when initial attempt fails. The second attempt is performed on reverse-complemented sequence.

If the flag is provided, as a consequence:
 - the output alignment, peptides and analysis results correspond to this modified sequence and not to the original
 - sequence name gets a suffix appended to it for all output files (fasta, `seqName` column, node name on the tree etc.)
 - in output files, there is a new field/column: `isReverseComplement`, which contains "true" if the corresponding sequence underwent reverse-complement transformation

This functionality is opt-in and the default behavior is unchanged: skip sequence and emit a warning.


rneher (approved):
This all works as expected for me.

PR #891, opened 2022-06-24
feat(cli): default basename to a consistent hardcoded value

ivan-aksamentov:
Currently, if `--output-basename` is not provided, and the basename for files written to `--output-all` is the same is for input fasta. However, if multiple fasta files provided, the output basename switches to a hardcoded "nextclade" or "nextalign".

This is not something that other CLI tools typically do and might be confusing, especially for use-cases where a certain output filename is expected (i.e. in scripts and pipelines), especially when a number of input fasta files is not known in advance or if it changes between 2 runs.

This PR proposes to always use a hardcoded name for consistency, so that there is no surprise.


rneher (approved):
I think this makes sense. pulling from the first file seems too arbitrary, we can't use a file prefix when reading from stdin, etc...

PR #916, opened 2022-06-30
feat: remove reverse complement suffix from outputs except nuc fasta

ivan-aksamentov:
Followup of https://github.com/nextstrain/nextclade/pull/887

We reconsidered: we will only append suffix to sequence names in aligned fasta output. So this removes the suffix from all other places (incl. translation fastas).


rneher:
this works as expected for me. The only thing I would suggest to change is the warning/error msg. this is currently

`...and 'reverse complement' suffix will be added to sequence ID.`

maybe

`...and 'reverse complement' suffix will be added to fasta header line in the nucleotide alignment.`

rneher (approved):

Issue #896, opened 2022-06-27
ENH: Make unsupported browser warning "louder", block "run" button


corneliusroemer:
I think we should make the notice much "louder" though - not orange but red, make it bold, maybe not just at the top but as an overlay that needs to be clicked away?

Also, we should probably not even allow the user to proceed with "run", blocking the button.

Right now, one really needs to read that orange text at the top, that's the only warning. Otherwise it errors as before.

<img width="1669" alt="image" src="https://user-images.githubusercontent.com/25161793/175939272-f4708a51-0274-425c-a16f-dc878d64c485.png">

_Originally posted by @corneliusroemer in https://github.com/nextstrain/nextclade/issues/866#issuecomment-1167280818_

ivan-aksamentov:
We can make it louder, but I would not break the "Run" button. Browser detection is somewhat unreliable and there might be false positives, especially if new browsers, and different privacy extensions are involved. Also, Apple may eventually fix the problem. There is no point of not allowing users to try.

PR #928, opened 2022-07-07
feat(web): improve unsupported browser warning

ivan-aksamentov:
Resolves https://github.com/nextstrain/nextclade/issues/896

This makes the warning to appear in a modal, so that it's harder to miss. Also adds links to official websites of Chrome and Firefox.

<a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/9403403/177699693-fe9e0606-55bc-4b39-a9d2-e519afc3980b.png">

</img>
</a>

### DIFF
--- a/packages_rs/nextclade-web/src/components/Common/BrowserWarning.tsx
+++ b/packages_rs/nextclade-web/src/components/Common/BrowserWarning.tsx
@@ -1,14 +1,19 @@
-import React, { useMemo } from 'react'
-import { useTranslation } from 'react-i18next'
-import { UncontrolledAlert } from 'reactstrap'
 import Bowser from 'bowser'
+import React, { useCallback, useMemo, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { Button, Modal, ModalBody, ModalFooter } from 'reactstrap'
+import { ModalHeader } from 'src/components/Error/ErrorPopup'
+import { LinkExternal } from 'src/components/Link/LinkExternal'
 
 import { notUndefinedOrNull } from 'src/helpers/notUndefined'
 
 export function BrowserWarning() {
   const { t } = useTranslation()
 
-  const warningText = useMemo(() => {
+  const [isOpen, setIsOpen] = useState(true)
+  const dismiss = useCallback(() => setIsOpen(false), [setIsOpen])
+
+  const nameAndVersion = useMemo(() => {
     const browser = Bowser.getParser(window?.navigator?.userAgent)
     const isSupportedBrowser = browser.satisfies({
       chrome: '>60',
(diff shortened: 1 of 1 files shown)


corneliusroemer:
Love the new modal, looks really good for me

Tested and only fires on Safari.

Maybe we should add something along the lines of: if you're getting this on Chrome/Firefox, please open an issue here so we find out if it gets shown to the wrong people, just to be safe?

packages_rs/nextclade-web/src/components/Common/BrowserWarning.tsx
```diff
@@ -21,20 +26,37 @@ export function BrowserWarning() {
...
     return null
   }
 
   return (
-    <UncontrolledAlert color="warning" className="text-center m-0">
-      {warningText}
-    </UncontrolledAlert>
+    <Modal centered isOpen={isOpen} backdrop="static" toggle={dismiss} fade={false} size="lg">
+      <ModalHeader toggle={dismiss} tag="div">
+        <h4 className="text-center text-danger">{t('Unsupported browser')}</h4>
```
corneliusroemer, line 39:
```suggestion
        <h4 className="text-center text-danger">{t('Unsupported browser: Please use Chrome or Firefox')}</h4>
```

packages_rs/nextclade-web/src/components/Common/BrowserWarning.tsx
```diff
@@ -21,20 +26,37 @@ export function BrowserWarning() {
...
+
+      <ModalBody>
+        <p>{t('This browser version ({{nameAndVersion}}) is not supported.', { nameAndVersion })}</p>
+        <p>
+          {t('Nextclade works best in the latest version of ')}
+          <LinkExternal href="https://www.google.com/chrome/">{t('Chrome')}</LinkExternal>
+          <span>{t(' or ')}</span>
+          <LinkExternal href="https://www.mozilla.org/firefox/browsers/">{t('Firefox')}</LinkExternal>
+          <span>{t('.')}</span>
+        </p>
```
corneliusroemer, line 50:
```suggestion
        </p>
        <p>
          {t('In addition to Chrome and Firefox, Nextclade should work with any Chromium based browser, such as Edge, Brave, Opera etc.')}
        </p>
        <p>
          {t('If you are shown this modal even though you're using a Chromium-based or Firefox browser, please let us know by opening an issue ')}
          <LinkExternal href="https://github.com/nextstrain/nextclade/issues/new?assignees=&labels=t%3Abug%2C+help+wanted%2C+good+first+issue%2C+needs+triage&template=01_report_bug.md">{t('here')}</LinkExternal>
          <span>{t('.')}</span>
        </p>
```

corneliusroemer (changes requested):
What do you think of these suggested changes?

packages_rs/nextclade-web/src/components/Common/BrowserWarning.tsx
ivan-aksamentov: This will look very ugly on mobile and also repeats what's in the body.

packages_rs/nextclade-web/src/components/Common/BrowserWarning.tsx
ivan-aksamentov: Regarding browsers, I think this is unnecessary. We are not here to teach people browsers landscape or promote anything. We just need the app to be usable.

Those who know, they don't need to be told. Those who don't know will likely only be confused by these extra details.

And again, this will not fit well on mobile.

packages_rs/nextclade-web/src/components/Common/BrowserWarning.tsx
ivan-aksamentov: And this is the first thing all iOS users will see when they navigate from Twitter.

corneliusroemer: Hmm, right - we should add that iOS does not work at all, even with Firefox/Chrome

PR #930, opened 2022-07-10
Let user use relative url for nextclade datasets server.

xzhub:
when the DATA_FULL_DOMAIN is started with '/', the HTTP Origin will
be added in front of it to make an absolute url.

### DIFF
--- a/packages_rs/nextclade-web/src/io/fetchDatasetsIndex.ts
+++ b/packages_rs/nextclade-web/src/io/fetchDatasetsIndex.ts
@@ -6,7 +6,11 @@ import urljoin from 'url-join'
 import { Dataset, DatasetsIndexV2Json } from 'src/algorithms/types'
 import { axiosFetch } from 'src/io/axiosFetch'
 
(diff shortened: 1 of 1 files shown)


packages_rs/nextclade-web/src/io/fetchDatasetsIndex.ts
```diff
@@ -6,7 +6,11 @@ import urljoin from 'url-join'
 import { Dataset, DatasetsIndexV2Json } from 'src/algorithms/types'
 import { axiosFetch } from 'src/io/axiosFetch'
 
-const DATA_FULL_DOMAIN = process.env.DATA_FULL_DOMAIN ?? '/'
+var DATA_FULL_DOMAIN = process.env.DATA_FULL_DOMAIN ?? '/'
```
ivan-aksamentov, line 9: Let's change this `var` to `let`.

```suggestion
let DATA_FULL_DOMAIN = process.env.DATA_FULL_DOMAIN ?? '/'
```

xzhub:
Thanks for the revise!

PR #938, opened 2022-07-13
fix: crash when running flu yam

ivan-aksamentov:


corneliusroemer (approved):

corneliusroemer:
Looks good, shall we add that? Could be useful in the future, who knows, and shouldn't cost that much.

PR #939, opened 2022-07-13
feat: add "download dataset and run" shortcut"

ivan-aksamentov:
This adds `--dataset-name` (`-d`) to `run` command, which allows to download a dataset with default parameters and run with it immediately, in one command.

For example this command.

```bash
nextclade run --output-all=out --dataset-name=sars-cov-2 sequences.fasta
```

will download the latest default sars-cov-2 dataset into memory and will run analysis with these dataset files.

This is a convenience shortcut for the usual combination of `dataset get` + `run`. The dataset is not persisted on disk and downloaded on every run.

This also adds `--server` flag to the `run` command, which works exactly the same as in the `dataset get` and `dataset list` commands.

### DIFF
--- a/packages_rs/nextclade-cli/src/cli/nextalign_loop.rs
+++ b/packages_rs/nextclade-cli/src/cli/nextalign_loop.rs
@@ -62,7 +62,7 @@ pub fn nextalign_run(run_args: NextalignRunArgs) -> Result<(), Report> {
   let gene_map = match input_gene_map {
     Some(input_gene_map) => {
       let gene_map = read_gff3_file(&input_gene_map)?;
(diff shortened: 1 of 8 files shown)


corneliusroemer:
Works well for me, would merge and make it a minor release once we're ready to release.

This is super neat for quick reproduction. The only thing that's missing for self-contained checks is that right now the command can't use example sequences that are downloaded.

For demo-mode and end-to-end tests we could  add a flag that uses sequences from the dataset itself, e.g. `--demo` 

What do you think @ivan-aksamentov ? Too specific use case to be worth implementing?

ivan-aksamentov:
I'll merge this and we can think of sample sequences separately. It is currently handled in different place in code.

PR #946, opened 2022-07-26
fix: use indices to identify sequences uniquely

ivan-aksamentov:
Nextclade is using sequence names to uniquely identify sequences. However, sequence names come from user inputs and cannot be trusted to be unique. Neither it seems there is a consensus on uniqueness of sequence names in the bioinformatics community as a whole.

This causes various problems where sequence names are used as identifiers, and when for example there are multiple sequences with the same name. In particular, when Nextclade Web stores analysis results, they are effectively stored in an associative container, where sequence name acts as a key. This leads to newer results overwriting older results as they arrive during analysis. Additionally, some of the HTML `id` properties used sequence names to add uniqueness. This was leading to incorrect HTML being produces, with multiple elements having the same `id` property.

In this PR I:
 - change internal storage to use sequence indices in the input file(s) as keys
 - add sequence index into HTML `id`s
 - display "Sequence index" in places where only sequence name was displayed previously

This should ensure correct handling of duplicated names.

This affects only web application. In the algorithmic and CLI parts, sequence names are not used - results are stored in the form of an array, and no HTML is involved.

### DIFF
--- a/packages_rs/nextclade-web/src/algorithms/types.ts
+++ b/packages_rs/nextclade-web/src/algorithms/types.ts
@@ -345,6 +345,7 @@ export function convertPrivateMutations(privateNucMutations: PrivateMutations) {
 }
 
 export interface AnalysisResult {
+  index: number
   seqName: string
   substitutions: NucleotideSubstitution[]
   totalSubstitutions: number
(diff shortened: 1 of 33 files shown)


corneliusroemer:
This seems to work well. A few comments:

1. I don't think we should present sequence index as prominently, if anything, add it behind the name, maybe right-aligned with a hash in front to show that it's sequence no. X, but not as a separate row that increases information overload
image


2. In the tree, duplicated sequences get a hash attached - not sure it's super important to align web table and tree, but this is an inconsistency.
image
image

ivan-aksamentov:
I removed indices for now.

PR #948, opened 2022-07-28
feat(web): report duplicate sequence names

ivan-aksamentov:
Followup of https://github.com/nextstrain/nextclade/pull/946

This adds little yellow icons in the "Sequence name" column when a sequence has the same name as other sequences in the same run. Indices of these sequences are additionally listed in the tooltip.


corneliusroemer:
That's great! Addresses my points from the original PR 
I haven't tested how this icon interacts with other icons that may pop up but should be fine - we can fix if there are issues.
This is definitely a nice and quick way to surface duplicates. Well done!

PR #956, opened 2022-08-09
feat: add coverage qc metric

### LINKED ISSUE #932: ENH: Add coverage to Nextclade output?
`Coverage` is normally defined as number of valid called bases divided by length of the virus.

Does Nextclade have a column for this?

The only way I can seem to compute it is to use
* totalMissing
* totalNonACGTNs
* alignmentStart
* alignmentEnd
along with the reference length.

But this seems error-prone and messy.

Is there an easier way?

ivan-aksamentov:
Resolves #932 
Related https://github.com/cevo-public/cov-spectrum-website/issues/545

In this PR:
 - [x] calculate coverage QC metric as defined in the issues linked above
 - [x] add it to CSV and TSV output
 - [x] add it to web UI ~- a circle with letter "c" in the "QC" column~ as a dedicated column


packages_rs/nextclade/src/run/nextclade_run_one.rs
```diff
@@ -149,12 +149,17 @@ pub fn nextclade_run_one(
 
   let aa_changes_groups = group_adjacent_aa_subs_and_dels(&aa_substitutions, &aa_deletions);
 
+  let total_aligned_nucs = alignment_end - alignment_start;
+  let total_covered_nucs = total_aligned_nucs - total_missing - total_non_acgtns - total_deletions + total_insertions;
+  let coverage = total_covered_nucs as f64 / ref_seq.len() as f64;
+
```
ivan-aksamentov, line 155: The formula needs scientific review and may need adjustments

packages_rs/nextclade/src/qc/qc_rule_coverage.rs
```diff
@@ -0,0 +1,35 @@
...
+  let score = clamp_min(score, 0.0);
+
+  let status = QcStatus::from_score(score);
+
+  Some(QcResultCoverage {
+    score,
+    status,
+    coverage,
+  })
+}
```
ivan-aksamentov, line 35: The logic for coverage QC score and status needs scientific review and adjustment

packages_rs/nextclade/src/qc/qc_config.rs
```diff
@@ -103,6 +103,25 @@ impl Default for QcRulesConfigStopCodons {
...
+impl Default for QcRulesConfigCoverage {
+  fn default() -> Self {
+    Self {
+      enabled: true,
+      score_bias: 0.0,
+      score_threshold: 100.0,
+    }
+  }
+}
+
```
ivan-aksamentov, line 124: Configuration part of the coverage QC metric is here. It is what's read from `qc.json` or substituted from the `Default` trait implementation. Then it is passed to `qc::qc_config::qc_rule_coverage::rule_coverage()` and can be used there to calculate QC score and status.

packages_rs/nextclade/src/io/nextclade_csv.rs
```diff
@@ -61,6 +61,9 @@ static NEXTCLADE_CSV_HEADERS: &[&str] = &[
   "alignmentScore",
   "alignmentStart",
   "alignmentEnd",
+  "qc.coverage.coverage",
+  "qc.coverage.score",
+  "qc.coverage.status",
```
ivan-aksamentov, line 66: The order of fields in CSV/TSV may need adjustment and can be changed here.

packages_rs/nextclade-web/src/components/Results/ListOfQcIsuues.tsx
```diff
@@ -61,6 +63,7 @@ export function ListOfQcIssues({ qc }: ListOfQcIssuesProps) {
     { name: t('Mutation Clusters'), shortName: 'C', value: snpClusters, message: formatQCSNPClusters(t, snpClusters) }, // prettier-ignore
     { name: t('Frame shifts'), shortName: 'F', value: frameShifts, message: formatQCFrameShifts(t, frameShifts) }, // prettier-ignore
     { name: t('Stop codons'), shortName: 'S', value: stopCodons, message: formatQCStopCodons(t, stopCodons) }, // prettier-ignore
+    { name: t('Coverage'), shortName: 'c', value: coverage, message: formatQCCoverage(t, coverage) }, // prettier-ignore
```
ivan-aksamentov, line 66: The letter inside the circle in the UI is defined here. Uppercase "C" was already used, so I figured lowercase "c" might be fine.

ivan-aksamentov:
We may or may not keep it is as a QC metric, depending on whether we want to signal low coverage or not.

If we don't want that, then we could just remove all the QC code here and move the `coverage` field into the main struct (which is reflected in JSON and NDJSON outputs) and rename the CSV column to simply `coverage`. In this case we need to decide where it should be displayed in the web UI.

corneliusroemer:
Fantastic Ivan, I will give this a deep check today.
You're a 🌟

packages_rs/nextclade/src/run/nextclade_run_one.rs
corneliusroemer: I think it's better to take insertions out - deletions are arguable. I'm not sure what bioinfo veterans would expect. Maybe @tseemann can help here?

Since one can simply delete insertions oneself we may want to go with this instead
```rust
let total_covered_nucs = total_aligned_nucs - total_missing - total_non_acgtns;
```

packages_rs/nextclade/src/qc/qc_rule_coverage.rs
corneliusroemer: Looks good, though we may want to have a max score as well? If people sequence only part of the genome on purpose we don't want to make their QC red. A max score of say 0.4 could help in this case.

packages_rs/nextclade/src/qc/qc_config.rs
corneliusroemer: Threshold needs to be on the order of 1 to produce scores between 0 and 100 - it's 100 times too large

corneliusroemer:
On way to surface coverage would be in the `missing` column - and/or the tooltip. After all it's related to missing, it's missing inside + missing at terminals.

packages_rs/nextclade-web/src/helpers/formatQCCoverage.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function formatQCCoverage<TFunction extends TFunctionInterface>(t: TFunction, qcCoverage?: QcResultCoverage) {
+  if (!qcCoverage || qcCoverage.status === QcStatus.good) {
+    return undefined
+  }
+
+  const { score, coverage } = qcCoverage
+
+  return t('Coverage is low: {{coverage}}. QC score: {{score}}', {
+    coverage,
```
corneliusroemer, line 13: Maybe express as `95%` rather than `0.95`?

packages_rs/nextclade/src/run/nextclade_run_one.rs
ivan-aksamentov: > alignment_end and alignment_start are positions on the reference, right?

Yes

> whether total_missing and total_non_acgtns count bases that are in insertions or not

No. Insertions historically are living their own life. They are stripped in Nextalign (called in the beginnign of this function) and we never touch them.

Might be something to discuss with Richard.

packages_rs/nextclade/src/qc/qc_rule_coverage.rs
ivan-aksamentov: Please do

packages_rs/nextclade/src/qc/qc_config.rs
ivan-aksamentov: Please adjust

ivan-aksamentov:
@corneliusroemer 

> On way to surface coverage would be in the missing column - and/or the tooltip

Are you saying that it's better to NOT make it a new QC metric and treat it as a simple number, like many other numbers in other columns?

packages_rs/nextclade-web/src/helpers/formatQCCoverage.ts
ivan-aksamentov: Sure, why not.

corneliusroemer:
I think it would be good to flag this as a moderate QC issue - but not cause a sequence to be red. That's why I suggested a max QC score above. But yeah, we should discuss with Richard. We can have both: treat it as a number in column and QC - the two are kind of separate (we treat missing both as QC issue and as a number to be shown).

rneher:
regarding the formula, I would simply report "what fraction of the reference do we have information for". And that would be
```
let total_covered_nucs = total_aligned_nucs - total_missing - total_non_acgtns;
```
assuming that all ambiguous nucleotides are in the aligned bits. That is almost always true, but not guaranteed.

rneher:
regarding how to surface: People have been confused that short sequences are labeled as shiny high quality ones. But it is not a quality issue per se. A little torn on this one. I'd rather not have it be a QC issue.

corneliusroemer:
>regarding how to surface: People have been confused that short sequences are labeled as shiny high quality ones. But it is not a quality issue per se. A little torn on this one. I'd rather not have it be a QC issue.

A compromise could be to have it show up as an orange badge but not contribute to the overall score. Then it would be easy to spot incomplete sequences quickly without having to parse numbers.

But in any case, if it's a column of its own people should be able to do their own filtering.

corneliusroemer:
So if we don't put it as a QC metric, we'd have to change the code a bit, right? @ivan-aksamentov

Surface it as a column of its own, just as a float.

ivan-aksamentov:
@corneliusroemer @rneher 

I removed the QC metric and added coverage as a column. 

Please:
 - check that all the outputs make sense (JSON, NDJSON, CSV, TSV, Web).
 - write tootltip text for the column:

https://github.com/nextstrain/nextclade/blob/ff9767b42f31bab460e686f746e15c9679603b5c/packages_rs/nextclade-web/src/components/Results/HelpTips/HelpTipsColumnCoverage.mdx?plain=1#L1-L3

ivan-aksamentov:
As an optional change, I also added new fields to output files

```
refLength
totalCoveredNucs
```

because this was also mentioned in the original issue. They are byproducts of calculation of coverage, so why not use them. I think some users may enjoy having ref length without needing to have the entire ref sequence around or hardcoding the length.

On the negative side: it's not that we don't have enough columns. Ref length is the same for all sequences and just duplicated.


Let me know if you want to keep that.

corneliusroemer:
Thanks that's great. I wouldn't yet include ref length and covered nucs - these are pretty easily derived as is.

We can always add them later but removing columns is worse.

corneliusroemer:
@ivan-aksamentov I'd get rid of the covered-nucs and ref-length columns for now and reduce the number of significant figures to 5 for the float of coverage. Also if we could move the column on the right of Ns that'd be great.

Otherwise ready to merge! Should be useful.

@chaoran-chen this should do the job, right? You get a number between 0 and 1 that says what proportion of the sequence we have information on.

packages_rs/nextclade-web/src/components/Results/ResultsTableRowResult.tsx
```diff
@@ -109,6 +110,10 @@ export function ResultsTableRowResult({
         </TableCellAlignedLeft>
       ))}
 
+      <TableCell basis={columnWidthsPx.coverage} grow={0} shrink={0}>
```
corneliusroemer, line 113: This needs to be moved, too - probably

corneliusroemer (changes requested):
Looks great, just a few tiny changes. I tested web and tsv/csv - all look good save the points I commented on.

packages_rs/nextclade-web/src/components/Results/ResultsTable.tsx
```diff
@@ -199,6 +250,16 @@ export function ResultsTable() {
             <HelpTipsColumnMissing />
           </ButtonHelpStyled>
         </TableHeaderCell>
+
```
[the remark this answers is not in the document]
ivan-aksamentov: Done.

Although I find that percentage column in between numeric columns looks a bit random. Also people will need to adjust their muscle memory to find the columns on the right side now.

packages_rs/nextclade/src/io/nextclade_csv.rs
```diff
@@ -170,6 +173,9 @@ impl<W: VecWriter> NextcladeResultsCsvWriter<W> {
       // private_aa_mutations,
       missing_genes,
       // divergence,
+      ref_length,
```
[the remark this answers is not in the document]
ivan-aksamentov: I reverted 44e90db56b01ae68448fefcc373fb88c3118c7e8

packages_rs/nextclade/src/io/nextclade_csv.rs
```diff
@@ -252,6 +258,9 @@ impl<W: VecWriter> NextcladeResultsCsvWriter<W> {
     self.add_entry("alignmentScore", &alignment_score)?;
     self.add_entry("alignmentStart", &alignment_start.to_string())?;
     self.add_entry("alignmentEnd", &alignment_end.to_string())?;
+    self.add_entry("refLength", ref_length)?;
+    self.add_entry("totalCoveredNucs", total_covered_nucs)?;
+    self.add_entry("coverage", coverage)?;
```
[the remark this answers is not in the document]
ivan-aksamentov: Why? That's just random. No other number in this output is formatted. Also, why 5 and not 3 or 6? 

This will only affect CSV and TSV and other outputs will be different. Makes no sense to me.

I think users should get the whole information and decide how many digits they want to take on their own.

PR #976, opened 2022-08-30
fix: attempt to fix old browsers [wip]

ivan-aksamentov:
Next.js 12 does not seem to be taking into account `browserslist` and won't transpile `Object.fromEntries` and spread operator `...`. In particular, Edge 18 and Firefox pre-65 don't understand the current bundle.

People have been complaining and `browserslist` support seems to be present in Next.js `12.2.6-canary.6`. However it also breaks a bunch of stuff. Let's monitor the situation.

https://github.com/vercel/next.js/discussions/33227
https://github.com/vercel/next.js/pull/36584


corneliusroemer:
Ah could this be behind some of the errors with Asian users?

ivan-aksamentov:
This is not related to regions, but yes, some of the error reports may be explained by this. There aren't many though, as not many people use pre-Chromium Edge or old Firefox (someone might get stuck at Firefox ESR 60 probably).

PR #987, opened 2022-09-08
fix: consider different file encodings

ivan-aksamentov:
This attempts to mitigate crash in web when the fasta file is not UTF-8 or ASCII encoded.

Here I am trying to deduce a list of encodings the file resembles and then attempting to decode the contents of the file from these encodings to a JS string, instead of simply using the `FileReader`'s `.toString()` function.


tsibley:
_taking a look due to a [Slack inquiry](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1662655539192319)_

> However, in my testing with fasta files, detection almost never works correctly.

Detection usually works by heuristics on the proportions of various bytes (often in combination with language word usage models). I suspect it fails here because the character range of FASTAs of genomes is going to be massively skewed due to the sequence content. So whatever the content is in the FASTA deflines (`>…`), it's likely to be considered noise or valid-but-mojibake bytes in some incorrect encoding. You really want a detector that operates only on the deflines themselves, but to do that you'd need to figure out how to isolate those. Perhaps using a fixed list of the `>` and `\n` markers in various encodings so you can trim the deflines out, or perhaps adjusting detection to ignore bytes decoding to the IUPAC alphabet range.

This sort of detection is also fraught with lots of false matches that decode "correctly" but not as intended. For example, windows-1251 (covering Cyrillic scripts) is often technically decode-able as windows-1252 (covering extended Latin scripts) but is gibberish. This is just one example of a great many of how it can go wrong.

image

Stepping back a bit, it seems like it'd be good to figure out if the actual errors users are getting is indeed encoding-related or not. If it is, then maybe more effort here is good, but unless confirmed to be encodings, it's still possible it's something else, right?

tsibley:
The JS `chardet` library also seems to get this specific example wrong where other detection libraries don't.

image

(`chardetect` is from Python's `chardet` library; `uchardet` is https://www.freedesktop.org/wiki/Software/uchardet/)

ivan-aksamentov:
@tsibley Thanks! I had similar thoughts. I found another chardet library that seems to be more accurate in detecting the encoding. I think we might deploy this and observe how this influences error rates and what detailed error messages we get.

My only worry for now is that to not break the main use-case, which is ASCII and UTF-8 encoded files.

corneliusroemer:
I could ask on Twitter whether anyone has had Nextclade crash and whether they would be so kind to send a sample file?

Or maybe someone reads our comments here and can get in touch? Would be good to get more issues from people for whom Nextclade crashes. We haven't received many - which is maybe a good sign, in that users realized what the problem was so they could fix it?

PR #990, opened 2022-09-12
feat: calculate escape

ivan-aksamentov:
Adds escape calculation.

Slack thread: https://neherlab.slack.com/archives/C015PFP5V44/p1662804325766339


packages_rs/nextclade/src/analyze/escape.rs
```diff
@@ -0,0 +1,22 @@
...
+      for (antibody, coefficient) in coefficients {
+        if let Some(weight) = escape_data.weights.get(antibody) {
+          escape += weight * coefficient;
+        }
+      }
+    }
+  }
+
+  escape
+}
```
ivan-aksamentov, line 22: The calculation is performed here, for one "gene" (or perhaps for an RBD, in case if it so happens that there are multiple RBDs).

packages_rs/nextclade-web/src/components/Results/HelpTips/HelpTipsColumnEscape.mdx
```diff
@@ -0,0 +1,3 @@
+##### Column: Escape
+
+Displays escape.
```
ivan-aksamentov, line 3: This is a content for the tooltip for the column in Web

PR #997, opened 2022-09-21
feat: make escape data amino acid based

corneliusroemer:


corneliusroemer:
Here's a virus properties json to use (txt ending so that github accepts it)

[virus_properties.json.txt](https://github.com/nextstrain/nextclade/files/9620182/virus_properties.json.txt)

ivan-aksamentov:
A similar feature is implemented as a part of https://github.com/nextstrain/nextclade/pull/990

Issue #998, opened 2022-09-22
ENH(web): Allow preselection of dataset configuration without immediate running


corneliusroemer:
It's amazing that one can pass dataset and sequences via URL parameters, this makes nice integrations possible for e.g. covSpectrum.

I often use a particular configuration of dataset that's always the same. Right now I can't preselect that by saving a URL. It'd be great if we could make the URL-param feature a bit more flexible to not require sequences to be passed.

Basically I want all input options to be set like one can through the UI - but not requiring the sequences yet, which I would then just drop.

For an example: I usually use `nextclade.dev` with the dataset `sars-cov-2-21L` and the reference tree downloaded from `https://api.codetabs.com/v1/proxy?quest=https://nextstrain.org/charon/getDataset?prefix=staging/nextclade/sars-cov-2/21L`

Right now, I always have to click to insert that tree link, but I don't see why we can't make this selection via URL when I open Nextclade, something like:
`master.clades.nextstrain.org?dataset=sars-cov-2-21L&input-tree=https://api.codetabs.com/v1/proxy?quest=https://nextstrain.org/charon/getDataset?prefix=staging/nextclade/sars-cov-2/21L` etc, of course properly URL encoded.

This would be super useful as it would allow Nextclade configuration sharing by link - making it less "clicky".

ivan-aksamentov:
@corneliusroemer I think it should mostly work in https://github.com/nextstrain/nextclade/pull/999.

Give it a try. Note that the correct URL param for setting a dataset is `dataset-name` (there is also `dataset-reference` and `dataset-tag`).

For your example tree, the URL will be:

https://nextclade-git-feat-web-url-read-dataset-witho-0511cb-nextstrain.vercel.app/?dataset-name=flu_yam_ha&input-tree=https%3A%2F%2Fapi.codetabs.com%2Fv1%2Fproxy%3Fquest%3Dhttps%3A%2F%2Fnextstrain.org%2Fcharon%2FgetDataset%3Fprefix%3Dstaging%2Fnextclade%2Fsars-cov-2%2F21L


The app initialization is quite complicated, there are many sources of incoming data and it is not easy to test. Would be nice if you could test that the normal flows did not break with this change.

PR #999, opened 2022-09-22
fix(web): read dataset name from URL, even when fasta is not provided

ivan-aksamentov:
Resolves: https://github.com/nextstrain/nextclade/issues/998

Now when `dataset-name` is present in the URL params, Nextclade will try to find and set this dataset as current, regardless of whether the `input-fasta` is also provided. This allows to set a dataset and customize its files using URL params, with intent to then load fasta manually.

Normally, when navigation happened from such URL, there is no obvious sign that dataset files were changed. I added a little indicator with a counter to make it more obvious when dataset files are customized.

Example: 

https://nextclade-git-feat-web-url-read-dataset-witho-0511cb-nextstrain.vercel.app/?dataset-name=sars-cov-2-21L&input-tree=https%3A%2F%2Fapi.codetabs.com%2Fv1%2Fproxy%3Fquest%3Dhttps%3A%2F%2Fnextstrain.org%2Fcharon%2FgetDataset%3Fprefix%3Dstaging%2Fnextclade%2Fsars-cov-2%2F21L

### DIFF
--- a/packages_rs/nextclade-web/src/components/Main/ButtonCustomize.tsx
+++ b/packages_rs/nextclade-web/src/components/Main/ButtonCustomize.tsx
@@ -6,6 +6,8 @@ import styled, { useTheme } from 'styled-components'
 
 import { useTranslationSafe } from 'src/helpers/useTranslationSafe'
 import { ButtonTransparent } from 'src/components/Common/ButtonTransparent'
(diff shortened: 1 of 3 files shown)


corneliusroemer:
Tested and it works - let's merge and fix if it cause problems on master.

I like the viz :) 

image

corneliusroemer (approved):

PR #1006, opened 2022-09-29
feat: allow hiding custom clade columns

ivan-aksamentov:
Related: https://github.com/nextstrain/nextclade_data/issues/39

Allows hiding custom clade columns from Nextclade Web.

In order to hide a column from Nextclade Web, set `"hideInWeb": true` in `meta.extensions.nextclade.clade_node_attrs` in reference tree JSON, for example:

```json
"extensions": {
  "nextclade": {
    "clade_node_attrs": [
      {
        "name": "WHO_name",
        "displayName": "...",
        "description": "...",
        "hideInWeb": true
      }
    ]
  }
}
```

The columns/fields will still be rendered in all other outputs: TSV, CSV, JSON and NDJSON

### DIFF
--- a/packages_rs/nextclade-web/src/state/results.state.ts
+++ b/packages_rs/nextclade-web/src/state/results.state.ts
@@ -215,9 +215,23 @@ export const hasTreeAtom = selector<boolean>({
   },
 })
 
-export const cladeNodeAttrDescsAtom = atom<CladeNodeAttrDesc[]>({
-  key: 'cladeNodeAttrDescs',
+const cladeNodeAttrDescsStorageAtom = atom<CladeNodeAttrDesc[]>({
+  key: 'cladeNodeAttrDescsStorage',
   default: [],
+})
+
+export const cladeNodeAttrDescsAtom = selector<CladeNodeAttrDesc[]>({
+  key: 'cladeNodeAttrDescs',
+  get({ get }): CladeNodeAttrDesc[] {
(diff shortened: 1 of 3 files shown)


corneliusroemer:
The default seems to be wrong - and set to "true"? I don't see Nextclade_pango despite it not being set at all

image

corneliusroemer:
I can only get it to either hide all or none - but not hide one and show one.

This is a test tree to use that should show Nextclade Pango, but not unaliased:
[hidepartially.json.txt](https://github.com/nextstrain/nextclade/files/9701433/hidepartially.json.txt)

corneliusroemer (changes requested):
Doesn't seem to work for me, see comment with test data - what am I doing wrong?

ivan-aksamentov:
@corneliusroemer There seems to be a bug - first run works as expected, but subsequent runs are not. Workaround is to reload the page. I'll investigate.

ivan-aksamentov:
@corneliusroemer I think [`d30d608` (#1006)](https://github.com/nextstrain/nextclade/pull/1006/commits/d30d60885f8a68d382b04570f388cdeb094503aa) should fix that. Please give another try.

ivan-aksamentov:
Something weird is still happening, only the values are hidden now, but not the whole column. I will investigate tomorrow.

corneliusroemer:
Oh god, sorry! Thanks for noticing.

corneliusroemer (approved):
It works 🎉

PR #1008, opened 2022-10-03
feat: calculate composite fitness

ivan-aksamentov:


corneliusroemer:
Excellent! Would it be possible to make it sortable by that column, too?
image

Issue #1022, opened 2022-10-14
Lineage XAN* (Nextclade) seems to include too many sequences?


Mike-Honey:
Cross posting from:
https://github.com/cevo-public/cov-spectrum-website/issues/614#issuecomment-1277235824

I noticed the XAN* (Nextclade) lineage is reporting 37K+ sequences, which seems too high.
https://cov-spectrum.org/explore/World/AllSamples/Past6M/variants?nextcladePangoLineage=XAN*

From the pango designation discussion, I was expecting under 100:
https://github.com/cov-lineages/pango-designation/issues/771

Searching for the GISAID query: Spike_F486V, NSP3_L1752M, NSP13_P529L, NSP4_L438F gave me 74.

Perhaps only the Spike mutation is being filtered?

Issue #1002, opened 2022-09-26
nextclade on S gene -  recombinant clade ?


JeromeAmbroise:
Hello,

When I try to apply nextclade (cli) on the s gene part of the genome, I obtain a lot of results associated 'recombinant' clade identification.
**One example:**
I downloaded the virus OM263444.1 |Severe acute respiratory syndrome coronavirus 2 isolate SARS-CoV-2/human/USA/0-56447-4/2021, complete genome from ncbi.
I applied nextclade on the full genome and I obtained (as expected) the clade 'BA.1'.
Then I extracted the gene s from the genome and I ran nextclade with the '--genes=S' option.
I obtained the following result for the clade : "recombinant".

I would be interested to know why I obtain this result .. and if there is a solution to obtain the clade 'BA.1' attribution based on the gene S only.

Thank you in advance

Jérôme

corneliusroemer:
Hi Jérôme,

It's a known issue that short sequences can get incorrectly called as recombinants.

The way to work around this is to use the "no-recombinants" dataset: `sars-cov-2-no-recomb` that is identical to the standard dataset but lacks all recombinants.

<img width="1707" alt="image" src="https://user-images.githubusercontent.com/25161793/192366526-cf3dc017-45a7-4d58-9566-a5f5ae2d8ec9.png">

`nextclade run -d sars-cov-2-no-recomb ...`

<img width="918" alt="image" src="https://user-images.githubusercontent.com/25161793/192366611-2c607da6-562f-4542-8857-07c91f6f0ae1.png">

Recombinants with breakpoint outside spike are indistinguishable to the parent that donated the spike when you only have spike present.

The long term solution is for us to bias against assigning recombinants - when a non-recombinant fits just as well.

Issue #918, opened 2022-06-30
Is it by design that sequence with lots of Ns are classified as recombinant?


danrlu:
This sequence has mostly Ns with 1 mutation and is "recombinant". Just checking that this is intended behavior. Certainly a little confusing.  
<img width="788" alt="image" src="https://user-images.githubusercontent.com/20667188/176790614-b69caf68-6ae7-4b04-a027-d464874b0c77.png">

corneliusroemer:
Hi Dan!

A sequence with 28816 Ns is very much a total edge case. How many non-N bases do you even have? It can't be more than a thousand!

In that regime, it's basically impossible to place sequences phylogenetically with any confidence - too little information.

The reason this ends up as a recombinant is that recombinants have stretches from two lineages, so in order to say whether something is a recombinant or not you need non-N bases from either side of the breakpoint. Otherwise it's impossible to be sure and one needs to use some sort of prior (which in this case would be very much against this sequence being a recombinant).

But we don't do any prior weighting - we just traverse the tree and choose the best match - maybe we could bias against assigning a recombinant if there's a non-recombinant that has the same parsimony score.

Otherwise, you could run it through the dataset that has no recombinants:
<img width="846" alt="image" src="https://user-images.githubusercontent.com/25161793/176895252-ae3c4bef-2b4d-4f68-a5cc-f605a6b9cb8d.png">

I hope that helps. There are two potential improvements we can make:
1. Improve the lineage we assign when there are equally parsimonious placements, e.g. bias against recombinants
2. Add a note of the number of equally parsimonious placements to show uncertainty

One question for you: did you notice the high number of Ns? The sequence is colored red so one shouldn't really trust the lineage assignment. Should that be made clearer?

danrlu:
This all makes sense. Very interesting to think through! 

This came up as we were QC-ing the sequence, and usually we would discard anything with so many Ns, but seeing it as recombinant made us think it twice and wondering whether we would miss anything interesting by throwing it away... But as you said the lineage assignment is not meaningful in this case. I remember there are other reasons a sequence would be red, right? Like too many mixed sites, I never ignore sequences b/c they're red. I think for seq w too many Ns, it's best to not show lineage at all. 

I think in general it could be helpful to show confidence in the lineage assignment one way or another, but then the question become whether that will be additionally confusing, and that's a bigger Q than this issue at hand.

Thanks very much for the great tool and always improving it!

Issue #761, opened 2022-03-28
ENH: Make it clearer when Pango calls are expected to be wrong


corneliusroemer:
It doesn't seem entirely clear to users when Pango assignments are trustworthy and when not.

One could for example add parentheses about the Pango assignment when the quality is bad to make the reduced confidence clear.

Or maybe the tooltip needs to be updated to give explicit hints on when to trust it and importantly when not:

https://github.com/nextstrain/nextclade/issues/760

AnonymousUserUse:
Is this issue already tackled, and if yes, is it visible in CoV-Spectrum?

I wonder whether the assignment of XAP from pangolin (80 sequences with coverage >0.9) or nextclade (1811 sequences with coverage >0.9) is more correct and hope that I can see the answer for those kinds of questions in certain website.

PR #1038, opened 2022-10-25
feat: use placement_priors as tie-breaker for placement

### LINKED ISSUE #605: Spike only sequences from VOC alpha aren't assigned the 20I (Alpha) but 20B
The presence of recombinant lineage `XA` with Alpha spike in a non-alpha background attracts Spike only Alpha sequences. 

Unsure how to handle this, but as a stop-gap solution we could remove recombinant lineages from the reference tree.

corneliusroemer:
Resolves #605 #1022 #1002 #918 #571 #394 #760 
Related to #761

Right now, we find nearest neighbors simply by minimizing the number of mutations that differ. When there are multiple nodes that minimize that difference, we pick the first node in a breadth first search.

This has caused issues, since recombinants are attached only 2-3 edges in, while BA.5 are about 10 nodes deep. So if a recombinant is among the candidate nodes with minimal difference, recombinants are picked.

This is not good as recombinants are actually unlikely to be the most reasonable choice in case of a tie.

This PR adds a `placement_bias` node attribute to the `tree.json` that is equal to the share of sequences that belong to the `Nextclade_pango` lineage that this node belongs to (this could be improved on in the future, as it ignores knowledge we have about number of sequences attaching to a particular node, example Spike only BA.5 with S:1020S gets assigned BA.5.2.1 though we know it should most likely be BF.5).

The nearest node algorithm then maximizes that `placement_prior` in case of a tie between multiple nodes.

Since `placement_bias` is an optional node attribute, this is backwards compatible: new datasets still work with old nextclade, and new nextclade works with old datasets.

We could further improve placement by e.g. ignoring terminal 200nt (often missing and hence not well known on the reference tree).

A test tree with placement priors is available here: 
[auspice.json.txt](https://github.com/nextstrain/nextclade/files/9864488/auspice.json.txt) and can be tested on the PR build by attaching: `?dataset-name=sars-cov-2-21L&input-tree=https://api.codetabs.com/v1/proxy?quest=https://github.com/nextstrain/nextclade/files/9864488/auspice.json.txt` to the URL.

And can be tested: https://nextclade-git-placement-priors-nextstrain.vercel.app/?dataset-name=sars-cov-2-21L&input-tree=https://api.codetabs.com/v1/proxy?quest=https://github.com/nextstrain/nextclade/files/9864488/auspice.json.txt

Extensions to consider:
- Ignore terminal say 100nt in distance calculation
- Use priors calculated by number of good, complete sequences attaching at each node rather than based on lineage frequency at that node
- Calculate and output placement confidence: how many equally parsimonious placements were there
- Find most likely node attribute by integrating over all equally parsimonious placement, alternatively: confidence for the attribute at the chosen attachment node
- Output file in pangolin format for compatibility


packages_rs/nextclade/src/run/nextclade_run_one.rs
```diff
@@ -113,10 +116,12 @@ pub fn nextclade_run_one(
   let unknown_aa_ranges = find_aa_letter_ranges(&translations, Aa::X);
   let total_unknown_aa = unknown_aa_ranges.iter().map(|r| r.length).sum();
 
+  // TODO: Ignore ends for placement -> more robust
   let TreeFindNearestNodeOutput { node, distance } =
     tree_find_nearest_node(tree, &substitutions, &missing, &alignment_range);
   let nearest_node_id = node.tmp.id;
-  let clade = node.clade();
+
+  let clade = node.node_attrs.clade_membership.value.clone();
```
corneliusroemer, line 124:
```suggestion
  let clade = node.clade();
```

packages_rs/nextclade/src/run/nextclade_run_one.rs
```diff
@@ -113,9 +116,11 @@ pub fn nextclade_run_one(
   let unknown_aa_ranges = find_aa_letter_ranges(&translations, Aa::X);
   let total_unknown_aa = unknown_aa_ranges.iter().map(|r| r.length).sum();
 
+  // TODO: Ignore ends for placement -> more robust
   let TreeFindNearestNodeOutput { node, distance } =
     tree_find_nearest_node(tree, &substitutions, &missing, &alignment_range);
   let nearest_node_id = node.tmp.id;
+
```
corneliusroemer, line 123:
```suggestion
```

ivan-aksamentov:
I've implemented better alternatives in:

 - https://github.com/nextstrain/nextclade/pull/1039 (adds the new field in a type-safe way)
 - https://github.com/nextstrain/nextclade/pull/1040 (complete revamp of nearest node search, without hacking into distance calculation, with multiple candidates emitted and with confidence score)

Discussion is here:
https://neherlab.slack.com/archives/C015PFP5V44/p1666738207897999


Please review and decide which way you want to go.

PR #1040, opened 2022-10-26
feat: list all nearest node candidates and then find the best one

ivan-aksamentov:
Alternative to https://github.com/nextstrain/nextclade/pull/1038

Here we gather a list of all nearest nodes (nodes with the same, minimum distance), then further sort the list according to some other metric. The first node in the resulting list will be the one used for placement.

The list can be displayed along with the confidence of each element in the Web UI and in the output files.

### DIFF
--- a/packages_rs/nextclade/src/run/nextclade_run_one.rs
+++ b/packages_rs/nextclade/src/run/nextclade_run_one.rs
@@ -114,7 +114,11 @@ pub fn nextclade_run_one(
   let total_unknown_aa = unknown_aa_ranges.iter().map(|r| r.length).sum();
 
   let nearest_node_candidates = tree_find_nearest_nodes(tree, &substitutions, &missing, &alignment_range);
-  let TreeFindNearestNodeOutput { node, distance } = nearest_node_candidates[0];
+  let TreeFindNearestNodeOutput {
+    node,
+    distance,
+    confidence,
+  } = nearest_node_candidates[0];
 
   let nearest_node_id = node.tmp.id;
   let clade = node.clade();
--- a/packages_rs/nextclade/src/run/nextclade_run_one.rs
+++ b/packages_rs/nextclade/src/run/nextclade_run_one.rs
@@ -23,7 +23,7 @@ use crate::run::nextalign_run_one::nextalign_run_one;
 use crate::translate::frame_shifts_flatten::frame_shifts_flatten;
 use crate::translate::translate_genes::{Translation, TranslationMap};
 use crate::tree::tree::AuspiceTree;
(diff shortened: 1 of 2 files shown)


packages_rs/nextclade/src/tree/tree_find_nearest_node.rs
```diff
@@ -3,35 +3,67 @@ use crate::analyze::letter_ranges::NucRange;
...
+
+/// Reorder equivalently near nodes according to a metric, from the best to the worst.
+/// The first node in the resulting list will be considered "the most nearest" among equally nearest nodes
+/// and will be used for clade assignment.
+fn tree_sort_nearest_nodes<'node>(
+  nearest_node_candidates: &[TreeFindNearestNodeOutput<'node>],
+) -> Vec<TreeFindNearestNodeOutput<'node>> {
+  // TODO: actually reorder the list or create a new one, and calculate placement confidence for each node
+  nearest_node_candidates.to_vec()
 }
```
ivan-aksamentov, line 67: This function needs to be implemented

Here we receive all candidate nearest node, (all having equal distance, which is the minimum distance, and we need to order the list according to our needs.

packages_rs/nextclade/src/run/nextclade_run_one.rs
```diff
@@ -113,8 +113,13 @@ pub fn nextclade_run_one(
...
   let total_unknown_aa = unknown_aa_ranges.iter().map(|r| r.length).sum();
 
-  let TreeFindNearestNodeOutput { node, distance } =
-    tree_find_nearest_node(tree, &substitutions, &missing, &alignment_range);
+  let nearest_node_candidates = tree_find_nearest_nodes(tree, &substitutions, &missing, &alignment_range);
+  let TreeFindNearestNodeOutput {
+    node,
+    distance,
+    confidence,
+  } = nearest_node_candidates[0];
```
ivan-aksamentov, line 121: The first of the sorted candidate nearest nodes is then considered for placement and clade assignment.

Wytamma:
Hi @ivan-aksamentov and @corneliusroemer,

Is there a plan to implement this feature in full? I've been using the list of distance metics to create heatmaps for each Nextclade run. It would be great to include a matrix of the distances in the Nextclade output and include something like the attached heatmap in the Web UI. 

heatmap

corneliusroemer:
Great to hear your ideas @Wytamma!

Do I understand you correctly, that for each query sequence (user provided) you want a list/dict of distances from each node present in the reference tree.

Can you explain the uses you have in mind? I can't quite see the general user benefit.

Are you trying to use it to spot recombinants? If that was the goal, it may be best to work on a fork/extension that leverages the existing code base.

For it to be _really_ useful for recombinant spotting, one would want to show distances not with respect to the entire sequence but chunks.

PR #1047, opened 2022-11-09
Update developer-guide.md

Actat:
Hello! I tried Nextclade web on my private server.

I followed instructions from the Nextclade Web section but faced an error.
I found `.env` is needed not only for Nextclade CLI but also for Nextclade web.
It works in my environment now.

So, it would be better to add this cp command in this section.

Thank you!


ivan-aksamentov:
Hi @Actat,

Thank you! Let us know if you notice any other mistakes or if you have ideas about improving dev workflow. Contributions are very welcome!

PR #1051, opened 2022-11-15
Feat/tree builder

anna-parker:
### Issue
When sequences are added to a reference tree in `nextclade` they are added to the node which is closest to that sequence according to a distance metric. They are not compared to each other and if most of them are closest to the same node this can result in a large polytomy and additional information about the relationship between nodes is lost. Therefore, if more than two sequences are closest to the same node we would like to build a subtree of these nodes instead of just adding them all to the same node. 

### Steps

- [x] 1. After finding all the closest node in the reference tree for each sequence in the alignment (i.e. at position: https://github.com/nextstrain/nextclade/blob/2b0a4797933dba0c60bd2c6a91734bfbe175a6cc/packages_rs/nextclade-cli/src/cli/nextclade_loop.rs#L261) see if >=2 sequences should be attached to the same node `c`. Write a function that takes `outputs` and returns lists of sequences that should be attached to the same node. 

For each group of sequences that should be attached to the same node (this can be multithreaded if the nodes each sequence should be attached to are not next to each other) perform the following steps, otherwise just attach the sequence directly: 
- [x] 2. Create a distance matrix between each sequence that should be attached to the node `c` and the node `c` using the distance metric: https://github.com/anna-parker/nextclade/blob/a2f2166867f5e9ba56b74d9d134e34a558735f19/packages_rs/nextclade/src/tree/tree_builder.rs#L40 (potentially `c.anc` and `c.children` also need to be added to the distance metric `d` to improve the tree structure).
- [x] 3. Use the neighbor joining algorithm to create an initial tree structure: https://en.wikipedia.org/wiki/Neighbor_joining 
- create a `Q` matrix from `d`
- find the minimum of `Q` and join the corresponding sequences to one node
- recalculate `d` and `Q`
- repeat steps until all sequences have been joined together.
- [ ] 4. (Potentially use the Felsenstein pruning algorithm to find the optimal branch length and sub tree topology)
- [x] 5. Join the new subtree to the node `c` (or alternatively, if `c.anc`, `c` and `c.children` are part of the distance matrix join the subtree to `c.anc`, and append the children of each child in `c.children` to their parent). 
- [ ] 6. Add all node data to nodes

### Test

A very simple dataset can be found in `data/flu/h3n2/ha_small`, this will attach 5 sequences to a two-node tree (where one node is the reference node and all sequences have been mutated from the reference node). The attached nodes should form a tree of the format "(((Sequence_2:1, Sequence_3:1)Sequence_1:2, (Sequence_5:1, Sequence_6:2)Sequence_4:2),other_node)reference". `Sequence_1` has been removed.


corneliusroemer:
@anna-parker something in the unit tests seems to fail - that's why CI says ❌ 
Not super consequential but would maybe be nice to have CI be ✅ 
https://github.com/nextstrain/nextclade/actions/runs/3695002858/jobs/6256931877

ivan-aksamentov:
@corneliusroemer This is a draft PR and it will take a while. But I am sure @anna-parker will fix all the tests towards the end.


I glanced over the code and I thought I mention a couple of handy (but optional) things which may help Anna in this adventure:

 - [`dbg!()` macro](https://doc.rust-lang.org/std/macro.dbg.html). The `dbg(foo)` is a superpowered version of `println!("{foo}")` (or `println!("{}", foo)`) - it returns its value, so you can stick it into expressions: 

	 ```rust
	 # What is b and a before the call?
	 let a = b;
     do_things(&a);
	
	 # Easy
	 let a = dbg!(b);
     do_things(dbg!(&a));
	 ```
   and you can print multiple things much easier too: `dbg!(foo, bar, baz)`. Plus it prints file, line and names of variables, so you can keep track of execution. Don't forget to take `&` references for variables where relevant (compiler will tell you).

 - [`match` constructs](https://doc.rust-lang.org/book/ch06-02-match.html). They are a superpowered, type-safe `switch` or `if/else`. Especially helpful for dealing with `enums`, including `Option` and `Result`. Things like `is_some()` + `unwrap()`:
 
 	 ```rust
	  let nums = vec![1, 2, 3];
	  let mut x = 1; // Unnecessary mutability
	  if nums.last().is_some() {
	    x = *nums.last().unwrap(); // May crash in ugly way if the `if` condition changes in the future
	  } else {
	    x = 1;
	  }
	  // x can be mutated accidentally here
	```

	are more verbose and less safe than

 	 ```rust
	  let nums = vec![1, 2, 3];
	  let x = match nums.last() {
	    Some(last) => x,
	    None => 1
	  };
      // x is immutable
	 ```

     or you might even enjoy some more functional approach for this common pattern:

     ```
	 let x = nums.last().unwrap_or(&1);
     ```
     (I am learning these from code completion ;) )

 - [Clippy](https://doc.rust-lang.org/nightly/clippy/index.html) - a linter (static analysis) tool. We have it [configured](https://github.com/nextstrain/nextclade/blob/d23f4490dceb2a06bd19ad41e4573b2f05d890a8/.cargo/config.toml#L88-L255) in the project. You can run it with `cargo clippy` and it will analyze your code and tell you where it can be improved, or where bugs might be. In some cases it also offers concrete suggestions of how to change code fragment to better.

   For example it will suggest you to transform the `match` in the previous example into `.unwrap_or()`.    It's a very good co-pilot and a mentor. Give it a try!

   There should be some sort of integration with VSCode. In CLion it's built-in and you can enable it in `File | Settings | Languages & Frameworks | Rust | External Linters`. They will show you yellow squiggles and offer quick-fixes. But this is optional. If you don't want to bother, then I will fix it after you are done.

    You can always lookup the explanations by the name of the rule [here](https://rust-lang.github.io/rust-clippy/master/).
 
 - Rustfmt - formats your code automatically. Run `cargo fmt` before you commit the code and the diffs will be less "jumpy". There is also integration into text editors.
 

All of this is optional, but feel free to give it a try, in case it helps you.

anna-parker:
@ivan-aksamentov and @corneliusroemer thanks so much for the comments! Clippy and fmt is very useful - I will try to use them from now on!

anna-parker:
@ivan-aksamentov I just realized I do not get all the clippy warnings I see here now when the tests fail due to the `cli / run lints`  test (https://github.com/nextstrain/nextclade/actions/runs/3732642335/jobs/6332374150) when I run `cargo clippy` locally. Could it be that I do not have the same version of clippy installed as is running on the docker container? I am running `clippy 0.1.63 (4b91a6e 2022-08-08)` locally but can't find which version is being run on the docker container.

ivan-aksamentov:
@anna-parker This might be due to differences in rust toolchain version. Clippy is installed as a part of the toolchain (typically done with `rustup`). I recently [upgraded it to 1.65](https://github.com/nextstrain/nextclade/blob/1dcb313ffe2eb3576548a82f97a27b60043952f3/rust-toolchain.toml#L2), but your branch is not in sync with this change.


Your branch:

```
$ cargo-clippy --version
clippy 0.1.63 (4b91a6e 2022-08-08)

$ rustc --version
rustc 1.63.0 (4b91a6ea7 2022-08-08)

$ rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/user/.rustup
1.63.0-x86_64-unknown-linux-gnu (overridden by '/workdir/rust-toolchain.toml')
rustc 1.63.0 (4b91a6ea7 2022-08-08)

```

master branch:

```
$ cargo-clippy --version
clippy 0.1.65 (897e375 2022-11-02)

$ rustc --version
rustc 1.65.0 (897e37553 2022-11-02)

$ rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/user/.rustup
1.65.0-x86_64-unknown-linux-gnu (overridden by '/workdir/rust-toolchain.toml')
rustc 1.65.0 (897e37553 2022-11-02)
```

The exact command that's running on CI is:

```
cargo clippy --exclude=3rdparty -q --target-dir=.build/docker/ --all -- -Dwarnings
```


Rust is a young and fast moving language. They release new versions at a 10x rate of Python ([1](https://github.com/rust-lang/rust/releases), [2](https://blog.rust-lang.org/)). In fact they've just released 1.66. If you want, you can merge `origin/master` into the branch periodically. And `cargo` should the pick up the new version from `rust-toolchain.toml` the next time you run any command involving rust in the project directory. It will install and use the version specified in the toml file (which may be different from your "default" installed toolchain). See [docs](https://rust-lang.github.io/rustup/overrides.html). But you of course can also install it manually. See `rustup --help`.

In any case - not a big deal. You don't have to maintain CI green at all times. The only requirement is that our final release is green. And I will definitely help.

So only care about clippy if it helps you to be more productive and do the thing you do better. And when it doesn't - you can pretty much ignore it. Right now the goal number one is to experiment, prototype and get the science right.

Ping me anytime here or on Slack.

PR #1076, opened 2022-12-20
Update Docker Hub login

victorlin:
Use new token and remove unnecessary encryption of username.

Part of org-level work in https://github.com/nextstrain/private/issues/73.


victorlin:
Merging without review since this is a trivial change and checks passed, which confirms that the new token works.

PR #1080, opened 2023-01-02
feat: find aa motifs

ivan-aksamentov:
Adds an additional analysis step: aminoacid motif search in translated and insertion-stripped peptides.

This is useful for example for detecting flu glycosylation spots.

### Configuration

Example config in `virus_properties.json`:

```json
  "aaMotifs": [
    {
      "name": "glycosylation",
      "nameShort": "Glyc.",
      "nameFriendly": "Glycosylation",
      "description": "Glycosylation motifs",
      "motifs": [
        "N[^P][ST][^P]"
      ]
    }
  ]
```
(also on [`glyc` branch](https://github.com/nextstrain/nextclade_data/blob/glyc/data/datasets/flu_h1n1pdm_ha/references/CY121680/versions/2022-12-07T08:35:53Z/files/virus_properties.json) of the data repo. The data there is currently random, do not merge!)

If `includeGenes` property is omitted, then all genes are searched in full. Otherwise, specified genes are searched. Search for a given gene can be restricted to a given codon range. If `ranges` are omitted, the whole gene is searched.

The `motifs` is an array of strings each containing a regular expression to search (in [Rust's regex crate syntax](https://docs.rs/regex/latest/regex/)).

The `name` is a unique identifier of the motif. The `friendlyName` is what's shown in web UI, and `shortName` is what's shown in narrow places of web UI (e.g. table column headers). The `description` is shown in the table column header tooltip.

### Results

Results are emitted into:
 - in JSON and NDJSON: property `motifs`, grouped by motif `name`
 - in CSV and TSV: in the column with name taken from motif's `name`; each cell containing a ';'-delimited list in format `{gene}:{position}:{motif_found}`.
 - in web UI table: a column name taken from motif's `name`, with total count of motifs in the cell and a table of motifs in its tooltip.

Web example:

https://nextclade-git-feat-motif-search-nextstrain.vercel.app/?dataset-url=gh:nextstrain/nextclade_data@glyc@/data/datasets/flu_h1n1pdm_ha/references/CY121680/versions/2022-12-07T08:35:53Z/files


packages_rs/nextclade-web/src/components/Results/ColumnAaMotifs.tsx
```diff
@@ -0,0 +1,349 @@
...
+
+  const preserved = useMemo(
+    () =>
+      motifs.preserved.length > 0 && (
+        <div>
+          <ColoredH6 className="mb-0">
+            {t('Preserved: {{preserved}}', { preserved: motifs.preserved.length })}
+          </ColoredH6>
+          <p className="my-0">
+            <small>{t('Motifs carried from reference sequence (sometimes mutated)')}</small>
```
rneher, line 96: Comment maybe 
"Present in reference and query (sometimes mutated, showing query motif)"

packages_rs/nextclade/src/analyze/find_aa_motifs.rs
```diff
@@ -0,0 +1,165 @@
...
+  seq: &str,
+  motif: &str,
+) -> Vec<Result<AaMotif, Report>> {
+  let re = Regex::new(motif)
+    .wrap_err_with(|| eyre!("When compiling motif RegEx '{}'", motif))
+    .unwrap();
+
+  re.captures_iter(seq)
+    .filter_map(|captures| {
+      captures.get(0).map(|capture| {
```
rneher, line 127: not 100% sure what is going on with `get(0)` here.

packages_rs/nextclade/src/analyze/find_aa_motifs_changes.rs
```diff
@@ -0,0 +1,153 @@
...
+        gene: motif_ref.gene,
+        position: motif_ref.position,
+        ref_seq: motif_ref.seq,
+        qry_seq: motif_qry.seq,
+      })
+      .sorted()
+      .collect_vec()
+  };
+
+  let total = gained.len() + lost.len() + preserved.len();
```
rneher, line 94: I think the `total` should be the just those in the query. that is `gained + preserved`.

rneher (commented):
In general this looks really good to me. A few minor comments (and my not understanding some syntax)

rneher:
I cross-checked a number of gains and losses with the associated mutations and it all checks out. 

Turns out the actual motif is just `N[^P][ST]` without the need for the extra `[^P]`

packages_rs/nextclade-web/src/components/Results/ColumnAaMotifs.tsx
ivan-aksamentov: It shows `ref -> qry` when mutated. If it shows only one, then it is not. This is why I splitted up mutated into a separate table initially.

packages_rs/nextclade/src/analyze/find_aa_motifs.rs
ivan-aksamentov: It takes the captures for the first capture group in the regex, which is the whole regex.

packages_rs/nextclade/src/analyze/find_aa_motifs_changes.rs
ivan-aksamentov: Fixed in f4c598b2b19c46411f7c3afc5b7cde363644e468

rneher (approved):

ivan-aksamentov:
@rneher Should we delay until there's a real-world dataset then? Just in case we need to change something.

rneher:
I do have a few fixes ....

if a gene is absent, it reports and error. 
image
Instead it should probably just ignore everything that is outside the alignment range. 

Same for partial genes:
image

here the gaps are outside the alignment range. 

All of this is using

https://nextclade-git-feat-motif-search-nextstrain.vercel.app/?dataset-url=gh:nextstrain/nextclade_data@glyc@/data/datasets/rsv_a/references/EPI_ISL_412866/versions/2022-12-20T22:00:12Z/files/

rneher:
If we have unknown amino acids, we maybe shouldn't report a missing motif as `lost`. Here a category `ambiguous` might be good:
image

from here: 
https://nextclade-git-feat-motif-search-nextstrain.vercel.app/?dataset-url=gh:nextstrain/nextclade_data@glyc@/data/datasets/sars-cov-2/references/MN908947/versions/2022-12-14T12:00:00Z/files

rneher:
something went wrong here: 
image

genes are no longer extracted correctly. It looks to me as if not the correct coordinate map is used.

ivan-aksamentov:
@rneher I added a few things to address the issues above:

 - [x] if a gene is missing, it will now silently ignore all motifs in this gene, instead of throwing an error
 - [x] calculate aa alignment ranges per gene. Please double-check the logic: src
 - [x] filter-out motif changes that are outside of aa alignment range
 - [x] display unsequenced regions in aa views in web, similarly to how it's done in nuc views.
 - [x] add "ambiguous" group of motifs. These are the motifs, containing at least one "X" in query. Can be seen for example in sequence #46 (OQ024138) of RSV-A dataset.

Questions:
 - Should "ambiguous" motifs be counted towards "total"
 - Totals are now confusing: partial sequence can have 4 preserved and total 4, while full sequence just below can have 10 preserved and 10 total. This kinda breaks concept of "preservation" - making it look like the reference sequences are different. We might need to either display the information differently or thoroughly explain what's going on. What can be done here?

rneher:
thanks. this looks much better. 

Regarding the "preservation". My understanding is that we detect motifs in the query sequence. The number we detect is the total. For convenience and interpretation,  we break this total down into those that are preserved and gained and in addition report those that are lost relative to the reference *in the region sequenced*.

I would thus omit ambiguous ones from the total (we don't report amino acid mutations in that region either) and replace the blue +/-# just by a dark grey number indicating the number of motifs that are either ambiguous in the ref or query.

ivan-aksamentov:
@rneher 

 - [x] removed ambiguous from total count
 - [x] removed the plus-minus sign and made the text of ambiguous grey

rneher:
Looks great. One final thing: in the tsv/csv export, the position of the motifs seems zero-based. In the web UI, it is one-based. Our tables are otherwise one-based, so there needs to be a plus one.

PR #1081, opened 2023-01-09
Allow to configure emitted CSV and TSV columns

ivan-aksamentov:
NOTE: Based on top of #1080 to avoid extra merge conflicts. The #1080 should be merged first.

Allow to restrict columns written into tabular output files (CSV and TSV).

### In CLI
  
Add flag `--output-columns-selection`.

Should contain a comma-separated list of individual column names and/or column category names to include into both CSV and TSV outputs. If this flag is omitted, or if category 'all' is present in the list, then all other entries are ignored and all columns are written. Only valid together with one or multiple of flags: `--output-csv`, `--output-tsv`, `--output-all`.

Possible values for categories: all, general, ref-muts, priv-muts, errs-warns, qc, primers, dynamic
Possible values for individual columns: column names as in CSV and TSV files

### In Web

Click "Configure columns" in the "Downloads" dialog. The configuration dialog opens. User can select categories and individual columns using checkboxes. There are "Download CSV" and "Download TSV" buttons for convenience at the bottom (additionally to the old ones, in the main "Download" dialog)

The selection is remembered across runs and page reloads (in local storage).


rneher:
I continue to think this is really useful. In the dialog, I would maybe rename

"Reference mutations" -> "Mutations relative to reference sequence"
"Private mutations" -> "Mutations relative to nearest node (private mutations)"
"Dynamic columns" -> "Dataset specific columns"

packages_rs/nextclade-cli/src/cli/nextclade_cli.rs
```diff
@@ -494,10 +495,28 @@ pub struct NextcladeRunOutputArgs {
...
+  /// Only valid together with one or multiple of flags: `--output-csv`, `--output-tsv`, `--output-all`.
+  #[clap(
+    long,
+    short = 'C',
+    takes_value = true,
+    multiple_values = true,
+    use_value_delimiter = true
+  )]
+  #[clap(requires = "tabular_output")]
+  #[clap(possible_values=CSV_POSSIBLE_CATEGORIES.iter().map(String::as_str))]
```
corneliusroemer, line 517: We need to add all individual column names here, too

ivan-aksamentov:
@corneliusroemer Should be fixed in https://github.com/nextstrain/nextclade/pull/1087

packages_rs/nextclade-cli/src/cli/nextclade_cli.rs
corneliusroemer: See https://github.com/nextstrain/nextclade/tree/fix-individual-csv-columns for a (partial?) fix

PR #1082, opened 2023-01-09
Allow to add example sequences using URL param

ivan-aksamentov:
Add special handling of value `example` to `input-fasta` URL param in Nextclade Web.

When `input-fasta=example` param is added, Nextclade Web will automatically start the analysis of example sequences (just like when clicking "Load example" on main page). The dataset should be provided with

This is handy when testing and debugging things.

Can also be used to combine external fasta and the examples. In this case, add multiple `input-fasta` parameters, e.g.

```
?input-fasta=example&input-fasta=http://example.com/my.fasta&dataset-name=my-dataset-name
```


rneher (approved):
this looks useful, mostly for us.

Issue #1085, opened 2023-01-15
Nextclade: Add a PINNED column 1,2,3,4 to easily count sequences of a lineage after sorting for them


FedeGueli:
Hi thanks for your great work.
I would suggest if it is possible to add at the start of the table before ( = at the left of) the ID Column a PINNED row with fix numbers from 1 to X to easily count numbers of sequences after sorting for (Pango lineages or Unaliased or whatsoever).

THX
<img width="1385" alt="Schermata 2023-01-15 alle 07 46 50" src="https://user-images.githubusercontent.com/87669813/212527030-f71b5c8d-89c0-439f-8d89-d810605de4bb.png">

ivan-aksamentov:
@FedeGueli Hi Federico, 

This sounds useful.

Just to understand your use-case better, do you need to see some sort of an aggregate statistics, such as:

 - clade "A": 3 sequences
 - clade "B": 10 sequences
 - clade "C": 4 sequences
 - total: 17 sequences
?

FedeGueli:
Thanks Ivan! that sounds great! but if it is too time expending it is enough a fixed column with 1 2 3 4 where i can count number of sequences belonging to a sublineage after sorting for example for unaliased lineage:
1-10 BA.5.1 = 10
11-16 Ba.5.2 = 5 and so on.

ivan-aksamentov:
@FedeGueli I implemented a prototype in https://github.com/nextstrain/nextclade/pull/1092

A preview app can be tested here: https://nextclade-git-feat-web-row-index-nextstrain.vercel.app/

Is this what you wanted?

FedeGueli:
> @FedeGueli I implemented a prototype in #1092
> 
> A preview app can be tested here: https://nextclade-git-feat-web-row-index-nextstrain.vercel.app/
> 
> Is this what you wanted?

Perfect!! It is exactly what i was meaning!

ivan-aksamentov:
@FedeGueli In https://github.com/nextstrain/nextclade/pull/1096 I made a prototype of simple clade statistics. Is this something that can help your use-case?

Here is a preview app: https://nextclade-git-feat-web-clade-stats-nextstrain.vercel.app/
The layout is not perfect yet and there might be bugs.

The team is hesitant to add complex post-processing features like this, so I don't guarantee it appears in the main app. But if there's strong need, it will be easier to justify.

[...]

PR #1092, opened 2023-01-25
feat(web): add row index column to results table

ivan-aksamentov:
Resolves https://github.com/nextstrain/nextclade/issues/1085

Adds column with index of the row in the table. Useful for visual search and counting of sorted and filtered results.

Not to be confused with sequence index.

Row indices always start with 0 and sorted in ascending order, and do not change their position when sorting or filtering the results.

Indices are 0-based.

These indices are not a part of output files. Spreadsheet software like Excel usually already have rows numbered in this manner.

In this screenshot, the "Row index" column is denoted as "#". The cells in this column have the same background as table header row, to indicate special meaning, just like in Excel. I sorted rows by clade and selected rows with 22F. If you look at row indices, you can quickly see that there are 6 rows (11 - 6 + 1) without counting them manually.

### DIFF
--- a/packages_rs/nextclade-web/src/components/Results/HelpTips/HelpTipsColumnRowIndex.mdx
+++ b/packages_rs/nextclade-web/src/components/Results/HelpTips/HelpTipsColumnRowIndex.mdx
@@ -1,9 +1,9 @@
 ##### Column: row index
 
-Index of the row in the table. Useful for visual counting of sorted and filtered results.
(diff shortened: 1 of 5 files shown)


corneliusroemer:
I'd say we drop the other index column from the view - as it's really of no use other than if you give all your sequences the same name and want to tell them apart - but no one does that and if anyone really wants to know the how manyth sequence it was we put that info in the tooltip if you hover over the sequence name.

Issue #1107, opened 2023-02-28
Random "ERROR: failed to solve" errors in docker build steps on GitHub Actions 


ivan-aksamentov:
We are facing random "ERROR: failed to solve" errors in docker build steps on GitHub Actions. Examples:

https://github.com/nextstrain/nextclade/actions/runs/4119091514/jobs/7469300033

https://github.com/nextstrain/nextclade/actions/runs/4287980131/jobs/7469450196

It is not immediately clear what's causing them and there don't seem to be any pattern or trend.

Context:

 - dockerfile: [./docker/docker-dev.dockerfile](https://github.com/nextstrain/nextclade/blob/master/docker/docker-dev.dockerfile) 
 - the script which runs on CI: [./docker-dev](https://github.com/nextstrain/nextclade/blob/ffec757976639a5a503a8c5bfea059d53e9052d7/docker-dev#L374-L429)

PR #1108, opened 2023-02-28
chore: remove `DOCKER_BUILDKIT` var

ivan-aksamentov:
Resolves https://github.com/nextstrain/nextclade/issues/1107

(or rather attempts to)

We are using `docker buildx` command, which AFAIU implies `DOCKER_BUILDKIT=1`.  
But someone on the internet* suggested that removing it or setting `DOCKER_BUILDKIT=0` may resolve the errors we are having. I'd like to keep using buildkit though, so don't know if any of it is a good idea. But let's give it a try on a branch.

 \* - https://github.com/docker/buildx/issues/415, https://stackoverflow.com/questions/64221861/an-error-failed-to-solve-with-frontend-dockerfile-v0


ivan-aksamentov:
After 2 successful runs in a row, it looks like this might indeed help. Let's give it a go on master and see how it runs.

PR #1116, opened 2023-03-09
fix(web): length of 3' unsequenced aa ranges

ivan-aksamentov:
The UI calculations were using nuc length of a gene, where there should be aa length. This led to wrong right boundary on 3' unsequenced range displayed in peptide views (e.g. above 3000 for gene S in SC2, while the gene is only 1200-something long)

In this fix, I ensure that `geneLength` is counted in codons, not nucs.

Before:

image

After:


corneliusroemer:
Bug was introduced with the `unsequencedEnd` markers in v2.10.0: https://github.com/nextstrain/nextclade/pull/1080

Reason I didn't spot it earlier is because I usually use this branch build which is stuck back in 2.8.0: https://nextclade-git-feat-composite-fitness-nextstrain.vercel.app/?dataset-name=sars-cov-2-21L&input-tree=https://nextstrain.org/charon/getDataset?prefix=staging/nextclade/sars-cov-2/21L

PR #1117, opened 2023-03-09
fix: wrong end coords of unsequenced 3' end in peptide viewer

corneliusroemer:


corneliusroemer:
Closed in favor of #1116

PR #1114, opened 2023-03-08
feat/placement priors 3

corneliusroemer:
Work in progress, but continuous review/refactoring appreciated!

- feat: add placement bias
- feat: list all nearest node candidates and then find the best one
- feat: add placement confidence score
- Print nearest node candidates with confidence

PR #1118, opened 2023-03-09
feat: add placement bias

ivan-aksamentov:
This proposes refactoring bits for https://github.com/nextstrain/nextclade/pull/1114

 - [x] use `TreeNodeAttrF64` as a type of `placement_prior` field on the tree, which contains f64 value instead of string. This avoid fallible parsing step and the need for magic values when parsing fails. This breaks. The tree needs to be changed. We need to thing whether f64 values are allowed in the tree attributes (notably in augur checks and in JSON schema).
 - [x] set `placement_prior` to `None` on new nodes, instead of empty string, to avoid magic values.
 - [x] deduplicate confidence score calculation and extract it into a separate function. This is handy to clearly set apart the algorithmic bits from the boilerplate and also avoids the problems like changing things in one place, but forgetting to change them in the other place.
 - [x] use the built-in `f64::powf()` instead of `libm::exp10()`. I hope I understand how it works correctly (perhaps arguments need to be swapped). This removes the extra dependency.

### DIFF
--- a/packages_rs/nextclade/src/tree/tree.rs
+++ b/packages_rs/nextclade/src/tree/tree.rs
@@ -35,8 +35,20 @@ impl TreeNodeAttr {
 }
 
 #[derive(Clone, Serialize, Deserialize, Validate, Debug)]
-pub struct TreeNodeAttrInt {
+pub struct TreeNodeAttrF64 {
   pub value: f64,
+
+  #[serde(flatten)]
+  pub other: serde_json::Value,
+}
+
+impl TreeNodeAttrF64 {
+  pub fn new(value: f64) -> Self {
+    Self {
+      value,
+      other: serde_json::Value::default(),
+    }
+  }
 }
 
 #[derive(Clone, Serialize, Deserialize, Validate, Debug)]
@@ -68,7 +80,7 @@ pub struct TreeNodeAttrs {
   pub division: Option<TreeNodeAttr>,
 
   #[serde(skip_serializing_if = "Option::is_none")]
-  pub placement_prior: Option<TreeNodeAttr>,
+  pub placement_prior: Option<TreeNodeAttrF64>,
 
   #[serde(skip_serializing_if = "Option::is_none")]
   #[serde(rename = "Alignment")]
--- a/packages_rs/nextclade/src/tree/tree_attach_new_nodes.rs
+++ b/packages_rs/nextclade/src/tree/tree_attach_new_nodes.rs
@@ -107,7 +107,7 @@ fn add_child(node: &mut AuspiceTreeNode, result: &NextcladeOutputs) {
         region: Some(TreeNodeAttr::new(AUSPICE_UNKNOWN_VALUE)),
         country: Some(TreeNodeAttr::new(AUSPICE_UNKNOWN_VALUE)),
         division: Some(TreeNodeAttr::new(AUSPICE_UNKNOWN_VALUE)),
-        placement_prior: Some(TreeNodeAttr::new("")),
+        placement_prior: None,
         alignment: Some(TreeNodeAttr::new(&alignment)),
         missing: Some(TreeNodeAttr::new(&format_missings(&result.missing, ", "))),
         gaps: Some(TreeNodeAttr::new(&format_nuc_deletions(&result.deletions, ", "))),
--- a/packages_rs/nextclade/src/tree/tree_find_nearest_node.rs
+++ b/packages_rs/nextclade/src/tree/tree_find_nearest_node.rs
@@ -56,38 +56,34 @@ pub fn tree_find_nearest_nodes<'node>(
   }
 }
 
-/// Reorder equivalently near nodes according to a metric, from the best to the worst.
+/// Reorder equivalently near nodes according to confidence.
 /// The first node in the resulting list will be considered "the most nearest" among equally nearest nodes
 /// and will be used for clade assignment.
 fn tree_sort_nearest_nodes<'node>(
   nearest_node_candidates: &[TreeFindNearestNodeOutput<'node>],
 ) -> Vec<TreeFindNearestNodeOutput<'node>> {
-  // TODO: actually reorder the list or create a new one, and calculate placement confidence for each node
-  let confidence_sum = nearest_node_candidates
+  let confidence_sum: f64 = nearest_node_candidates
     .iter()
-    .map(|node| {
-      node
-        .node
-        .node_attrs
-        .placement_prior
-        .clone()
-        .map_or(0_f64, |attr| libm::exp10(attr.value.parse::<f64>().unwrap_or(-10_f64)))
-    })
-    .sum::<f64>();
+    .map(|output| calculate_confidence(output.node))
+    .sum();
+
   nearest_node_candidates
     .iter()
     .map(|node| TreeFindNearestNodeOutput {
       node: node.node,
       distance: node.distance,
-      confidence: node
-        .node
-        .node_attrs
-        .placement_prior.clone()
-        .map_or(0.0, |attr| libm::exp10(attr.value.parse::<f64>().unwrap_or(-10_f64)) / confidence_sum),
+      confidence: calculate_confidence(node.node) / confidence_sum,
     })
-    // Calculate sum of all confidences
     .sorted_by(|a, b| b.confidence.total_cmp(&a.confidence))
     .collect_vec()
+}
+
+fn calculate_confidence(node: &AuspiceTreeNode) -> f64 {
+  node
+    .node_attrs
+    .placement_prior
+    .as_ref()
+    .map_or(0.0, |attr| 10_f64.powf(attr.value))
 }
 
 /// Calculates distance metric between a given query sample and a tree node
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1474,7 +1474,6 @@ dependencies = [
  "indexmap",
  "itertools",
  "lazy_static",
- "libm",
  "log",
  "maplit",
  "num",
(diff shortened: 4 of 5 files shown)


packages_rs/nextclade/src/tree/tree.rs
```diff
@@ -35,8 +35,20 @@ impl TreeNodeAttr {
...
+}
+
+impl TreeNodeAttrF64 {
+  pub fn new(value: f64) -> Self {
+    Self {
+      value,
+      other: serde_json::Value::default(),
+    }
+  }
 }
```
ivan-aksamentov, line 52: The `TreeNodeAttrF64` type to be used for `placement_prior`. Just like `TreeNodeAttr`, but with f64 inside.

packages_rs/nextclade/src/tree/tree.rs
```diff
@@ -68,7 +80,7 @@ pub struct TreeNodeAttrs {
   pub division: Option<TreeNodeAttr>,
 
   #[serde(skip_serializing_if = "Option::is_none")]
-  pub placement_prior: Option<TreeNodeAttr>,
+  pub placement_prior: Option<TreeNodeAttrF64>,
 
   #[serde(skip_serializing_if = "Option::is_none")]
```
ivan-aksamentov, line 85: Use `TreeNodeAttrF64` for `placement_prior`

packages_rs/nextclade/src/tree/tree_find_nearest_node.rs
```diff
@@ -56,40 +56,36 @@ pub fn tree_find_nearest_nodes<'node>(
...
+fn calculate_confidence(node: &AuspiceTreeNode) -> f64 {
+  node
+    .node_attrs
+    .placement_prior
+    .as_ref()
+    .map_or(0.0, |attr| 10_f64.powf(attr.value))
+}
+
 /// Calculates distance metric between a given query sample and a tree node
 pub fn tree_calculate_node_distance(
```
ivan-aksamentov, line 90: Extract confidence calculation into a function and use it in 2 places.

packages_rs/nextclade/src/tree/tree_find_nearest_node.rs
```diff
@@ -56,40 +56,36 @@ pub fn tree_find_nearest_nodes<'node>(
...
 
+fn calculate_confidence(node: &AuspiceTreeNode) -> f64 {
+  node
+    .node_attrs
+    .placement_prior
+    .as_ref()
+    .map_or(0.0, |attr| 10_f64.powf(attr.value))
+}
+
 /// Calculates distance metric between a given query sample and a tree node
```
ivan-aksamentov, line 89: Extract confidence calculation into a function and use it in 2 places.

There is no more parsing step, so no magic value is needed.

Use `f64::powf()` instead of `libm::exp10()`.

packages_rs/nextclade/src/tree/tree.rs
corneliusroemer: Ah - got it, the Int was how you had suggested to use the priors before.

Does this F64 mean I can't wrap the value in quotes in JSON, or does this not matter?

Would both of these work for deserialization?
```
value: "3.123"
value: 3.123
```

corneliusroemer: The default should probably be NAN, so set it explicitly as I'm not sure default isn't 0.

packages_rs/nextclade/src/tree/tree_attach_new_nodes.rs
```diff
@@ -107,7 +107,7 @@ fn add_child(node: &mut AuspiceTreeNode, result: &NextcladeOutputs) {
         region: Some(TreeNodeAttr::new(AUSPICE_UNKNOWN_VALUE)),
         country: Some(TreeNodeAttr::new(AUSPICE_UNKNOWN_VALUE)),
         division: Some(TreeNodeAttr::new(AUSPICE_UNKNOWN_VALUE)),
-        placement_prior: Some(TreeNodeAttr::new("")),
+        placement_prior: None,
         alignment: Some(TreeNodeAttr::new(&alignment)),
         missing: Some(TreeNodeAttr::new(&format_missings(&result.missing, ", "))),
```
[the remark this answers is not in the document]
corneliusroemer: None means it won't appear in the tree at all - not as NAN, right? Float defines nan, hence the question.

packages_rs/nextclade/src/tree/tree_find_nearest_node.rs
```diff
@@ -56,40 +56,36 @@ pub fn tree_find_nearest_nodes<'node>(
...
-    .map(|node| {
-      node
-        .node
-        .node_attrs
-        .placement_prior
-        .clone()
-        .map_or(0_f64, |attr| libm::exp10(attr.value.parse::<f64>().unwrap_or(-10_f64)))
-    })
-    .sum::<f64>();
+    .map(|output| calculate_confidence(output.node))
```
corneliusroemer, line 67: Any idea how to make this even dryer? We iterate once just to get the sum - but at least your way is way dryer already :)

packages_rs/nextclade/src/tree/tree_find_nearest_node.rs
corneliusroemer: Yep - this requires careful thought - so that there are no science bugs in edge cases. Doesn't matter in well behaved situations but in edge cases. Maybe add a "TODO in future" or something.

packages_rs/nextclade/src/tree/tree.rs
ivan-aksamentov: Nope, the "3.123" is string. If the Rust type is `f64`, serde will expect the number (without quotation marks).

ivan-aksamentov: The default is `None`, and can be handled as such (with `Option::unwrap*()` or `Option::map*()` ). No need to invent magic/bogus values.

packages_rs/nextclade/src/tree/tree_attach_new_nodes.rs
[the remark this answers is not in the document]
ivan-aksamentov: The full type is

https://github.com/nextstrain/nextclade/blob/650a5eb4a9811d4c8b4bbb9020d7a6c0a5c4a92c/packages_rs/nextclade/src/tree/tree.rs#L83

The options can be set to `None` manually or by serde (in case the field is no present)

packages_rs/nextclade/src/tree/tree.rs
ivan-aksamentov: > it would be nice if we could deal with either situation

I don't see why it would be nice. It would be super ugly in my view. Why would you want to complicate things and do 2 useless manual conversions:

 - number to string in your Python code
 - string to number in Rust (and handle potential failures)

when a type system can handle that.

Number is a number. Not string. If it's a number, it's guaranteed to be there and can be used without errors (the parsing and type checking is handled by serde-json).

Just remove the useless conversion code in Python and everything will be working again (and if it requires rerunning the pipeline, for the time being just convert them manually with sed or something)

packages_rs/nextclade/src/tree/tree.rs
```diff
@@ -35,8 +35,20 @@ impl TreeNodeAttr {
...
+
+  #[serde(flatten)]
+  pub other: serde_json::Value,
+}
+
+impl TreeNodeAttrF64 {
+  pub fn new(value: f64) -> Self {
+    Self {
+      value,
+      other: serde_json::Value::default(),
```
corneliusroemer, line 49: What is the purpose of these `other`s in the `TreeNodeAttr`s? I don't think I've understood that yet

packages_rs/nextclade/src/tree/tree_attach_new_nodes.rs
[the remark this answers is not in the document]
ivan-aksamentov: Yes, it will not appear. This is controlled by the `skip_serializing_if` annotation:

https://github.com/nextstrain/nextclade/blob/650a5eb4a9811d4c8b4bbb9020d7a6c0a5c4a92c/packages_rs/nextclade/src/tree/tree.rs#L81-L83

P.S. the `Option` can be set to `None` manually or by serde (in case the field is no present)

packages_rs/nextclade/src/tree/tree.rs
ivan-aksamentov: I think that the object there can contain other fields (which we don't care about). This is to tell serde to potentially expect more fields and to gather them in this dict-like structure.

It's no different from the string-based counterpart of this type.

Anna found these fields useful for prototyping (you can add things to JSON without modifying the struct)

corneliusroemer: What is the purpose of these `other`s in the `TreeNodeAttr`s? I don't think I've understood that yet

ivan-aksamentov: I believe there can be attrs on the tree which look like:

```json
{
	"name": "Foo"
	"description": "Yep, that's a foo!"
    "value: "foo!"
}
```

When parsing, the JSON `value` will go to struct's `value`, while the JSON's `description` and `name` will go to the struct's `other` "dict".

It's not strictly necessary for the f64 variant, which we only use in a single place we ourselves invented, but I like to add it anyway, to avoid "unexpected extra field" errors in case this changes.

ivan-aksamentov (commented):
> fix: accept quoted float (json string) for placement prior

"fix" :)

Fantastic way to complicate your life :)

PR #1119, opened 2023-03-09
feat: use placement priors as tie breaker

corneliusroemer:
This feature improves the placement of sequences that
could be added to multiple nodes on the tree
with the same distance

As tie-braker we use "priors", a number that is present on every tree node

That number represent an a priori likelihood
that sequences attach to that node

The calculation details of this prior could change. All that matters for now
is the ordering of priors

The numerical value has no purpose (yet)
In the future, we intend to use the priors
to display placement confidence information.

This feature should be backwards compatible. Trees without priors should work just as before.

This commit is a placement-only subset of PR #1114

Priors need to be a float with path `node_attrs.placement_prior.value` - they should be on log10 scale (so -2 to mean 0.01).

For SARS-CoV-2, the priors are added in https://github.com/neherlab/nextclade_data_workflows/pull/38

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 ## Upcoming
 
+### Attach sequences to a priori most likely node if reference tree contains "placement_prior"
+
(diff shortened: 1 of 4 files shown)


ivan-aksamentov:
LGTM in terms of code, but I would postpone merge until some real data is available and after the sciency results are reviewed by Richard and/or other scientists. Otherwise it's just dead code on master that might potentially need to be changed.

PR #1122, opened 2023-03-13
feat/output nearest nodes

corneliusroemer:
- refactor: create vec of nearest nodes by distance
- feat: output vec of nearest node names into ndjson (unstable)
This feature is to help with creating priors for Nextclade ref trees
The output is limited to ndjson and should be considered unstable
It may be removed without notice - hence should only be used by developers

This feature is a requirement for https://github.com/neherlab/nextclade_data_workflows/pull/38

The new `nearest_nodes` ndjson only output is for developers' use only (unstable/temporary).


corneliusroemer:
Not sure whether this should be minor or patch release - since it's only internal/unstable feature it may as well be patch.

ivan-aksamentov:
I only glanced over and haven't tried to run.

I think versioning-wise we define it as a non-breaking feature (even if internal), so minor version. We say in the docs that the JSON outputs are not stable, so it should be considered non-breaking.

How much new data do you add? Is that `n_query_seq * n_tree_nodes` strings, each string containing a name of a node? The same JSON is used when wasm module talks to the web app, so it's not desirable to add lots of data there. JSON parsing hangs the main thread, and causes rendering gaps and blips, which you dislike. Did you test the web app? Is there any visible effect?

Are strings good enough to identify the nodes? We identify attachment nodes by node index (indexing happens in `tree_prepare*` and follows some tree traversal order). They should be available at the `node.tmp.something` on the Rust side. If you want, you could do the same indexing on Python side and match by index to make it more reliable.

corneliusroemer:
Thanks for the review @ivan-aksamentov!

> I only glanced over and haven't tried to run.
I checked CLI and web and found no issues at all.

> I think versioning-wise we define it as a non-breaking feature (even if internal), so minor version. We say in the docs that the JSON outputs are not stable, so it should be considered non-breaking.
Excellent

> How much new data do you add? Is that n_query_seq * n_tree_nodes strings, each string containing a name of a node? The same JSON is used when wasm module talks to the web app, so it's not desirable to add lots of data there. JSON parsing hangs the main thread, and causes rendering gaps and blips, which you dislike. Did you test the web app? Is there any visible effect?
It adds ~1-10 node names to each line of JSON output. Only for very bad sequences is it sometimes up to 100, but amortized it almost never ends up above 10. Compared to all the rest we output into JSON it's a tiny drop in a huge pond. If we are worried about edge cases we could limit to `n` characters and/or `n` nodes.

> Are strings good enough to identify the nodes? We identify attachment nodes by node index (indexing happens in tree_prepare* and follows some tree traversal order). They should be available at the node.tmp.something on the Rust side. If you want, you could do the same indexing on Python side and match by index to make it more reliable.
This is a good question. Names _should_ be unique but this may not always be guaranteed. I'll bear this in mind - if non-uniqueness occurs and causes issues I could switch to using integer indexes, also to save space.

This however comes at the cost of making the nodes less human interpretable and when prototyping it's super useful to know at a glance which node it is. Since this is (mostly) internal we should be able to change from name to index any time.

I take your safety points and merge anyways to move ahead 🙃 I hope that's ok.

ivan-aksamentov (approved):

PR #1123, opened 2023-03-14
feat: only emit nearest node info if a CLI flag is set

ivan-aksamentov:
Followup of https://github.com/nextstrain/nextclade/pull/1122

Introduces a CLI flag `--include-nearest-node-info` and the correponding internal parameter for Nextclade runs. Skips the `.nearestNodes` field in JSON/NDJSON unless this flag is present.

This parameter is always `false` in web, so this data is never included into the outputs of the web app. This also helps to slightly reduce the amount of data passed between webworkers


corneliusroemer (approved):
Tested and it works

Quite a bit of piping and adds to the args shown to users so maybe we can refactor these into a dev option at some point, or output verbosity (as opposed to log verbosity).

PR #1125, opened 2023-03-15
fix: ensure phenotype values are attached to new tree nodes

ivan-aksamentov:
This adds custom phenotype values (such as `ace2_binding` and `"immune_escape"`) to the newly placed tree nodes.

They were previously missing, even though the technically similar clade-like node attributes were set. Which makes it somewhat asymmetric. 

I think it was simply an omission when the phenotype values were added.

Here I add the `phenotype_values` to the node similarly to how the `custom_node_attributes` are added.

This allows to see the values and colorings for phenotype values (if they are present in the dataset) on the tree page:


corneliusroemer:
Now I get the meaning of the `other` in the `TreeNode` struct - it's for things like custom node attributes.

Makes sense to add these values - not super important but why not for symmetry.

corneliusroemer (approved):
Tested and works

PR #1127, opened 2023-03-19
fix: incorrect 0 coercion

ivan-aksamentov:
The `pos` of  `0` caused to show mutation badges with position 0, while they are supposed to be 1-indexed. This was due to incorrect conversion of `0` to `false`.

Here I fix the problem by using the explicit comparison to `undefined` and `null` (using the helper function `isNil()` from `lodash`), which avoids implicit conversions.


corneliusroemer:
works 
image

PR #1128, opened 2023-03-20
feat(tree): Add masked sites for distance calculation

corneliusroemer:
*Currently still work in progress, but early review encouraged @ivan-aksamentov*

Allow setting masked sites: sites that should be ignored for tree placement purposes

Nextclade currently has no concept of masking - because we don't want to hide information from users in sequence view and for QC purposes.

Certain sites (e.g. terminals) are often poorly sequenced - such sites should not be used for reference tree placement, as the reference tree itself is uncertain at these positions.

We could mask these sites on the reference tree, but then the user would not see any mutational differences. It is better if we just mask the sites for placement purposes (and potentially for QC calculation, to be considered in future PR).

This PR adds the ability to specify sites/ranges to be ignored for placement purposes.

The `tree_calculate_node_distance` function is extended to accept `masked_ranges`.

The implementation simply removes query mutations in these ranges and adds the ranges as unknowns.

The masked ranges are configured by the user through `virus_properties.json`. By default, nothing is masked. This ensures backwards compatibility.

For SARS-CoV-2, masked ranges could for example be `1-100` and `29710-29903`. But if there a jumpy sites in the interior, these could also be included.

Beyond more robust placement, this feature could also be leveraged to search for recombinants and their donors: e.g. search for the lineage(s) best matching a 5' segment (0-10k) and a 3' segment (25k-30k) - usually the best matches should overlap, if they don't that's indicative of recombination.

Test links:
- [Only using 100-10700 for placement (for drastic effect)](https://nextclade-git-feat-masked-sites-for-placement-nextstrain.vercel.app/?dataset-name=sars-cov-2&input-pcr-primers=https://gist.githubusercontent.com/corneliusroemer/2543424bc004f2bb3fa9410edd74c6fa/raw/531a3d91e3f79fc56d16549b7e1b7e680140b9c8/virus_properties.json)
- [Masking just the terminals (0-100 and 29720-29903), how we'd use it in practice](https://nextclade-git-feat-masked-sites-for-placement-nextstrain.vercel.app/?dataset-name=sars-cov-2&input-pcr-primers=https://gist.githubusercontent.com/corneliusroemer/2543424bc004f2bb3fa9410edd74c6fa/raw/531a3d91e3f79fc56d16549b7e1b7e680140b9c8/virus_properties.json)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## Upcoming
+## Nextclade Web 2.13.0, Nextclade CLI 2.13.0 (2023-03-21)
 
 ### Attach sequences to a priori most likely node if reference tree contains "placement_prior"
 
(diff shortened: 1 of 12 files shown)


ivan-aksamentov:
I only glanced over and it makes sense. But I don't have full understanding of the underlying science, so here is a couple of notes:
 - Adding masked ranges to missing ranges seems to be increasing the `undetermined_sites` score, and the final distance unexpectedly. My understanding is that all SC2 sequences will get a constant addend to the score, so the score being added is effectively eliminated. I did not understand this part.
 - Could it be that in the future you want to mask not just by position, but also depending on the letter in the sequence? In this case perhaps a list of concrete mutations to ignore would suite better. But that would not be practical if you want to ignore large ranges. Or we can have both somehow.

Nextclade is not SC2-only software, so we might want to think about how it would serve for the analysis of other pathogens as well.

corneliusroemer:
Proof of concept seems to work, when placing based on 25000-end, it places various recombinants based on that fragment. Most reversions are in the region that's masked for placement - as expected, since that part isn't minimized during placement.

corneliusroemer:
@ivan-aksamentov 

> Adding masked ranges to missing ranges seems to be increasing the undetermined_sites score, and the final distance unexpectedly. My understanding is that all SC2 sequences will get a constant addend to the score, so the score being added is effectively eliminated. I did not understand this part.

I think (hope) you are missing something here. The masked ranges act as if you have extra Ns in the query sequence. If you trust the distance calculation without mask, and trust the code changing the query substitutions (and missing) then you should trust the new function - that's the beauty of the way it's implemented. I thought of a few ways to do this, but this way is the easiest to understand. 

You can convince yourself by going through all the cases that the distance can only _decrease_ by using the masked ranges. 
To go through the cases:

Masking:
- mutation in qry, no mut in node: undetermined unchanged (as not mutated in node), total_seq_muts -1 => distance -1
- no mut in qry, mut in node: undetermined +1, total_seq_muts unchanged => distance -1
- no mut in qry, no mut in node: everything unchanged
- mut in qry, same mut in node: shared difference -1, undetermined site +1, total  seq muts -1 => distance: -1 - 2 * (-1 ) - 1 = 0 -> distance unchanged as expected
- mut in qry, different mut in node: shared sites -1 (because mut removed from qry muts), undetermined +1, total seq muts -1 => distance -1 - (-1) - 1 -> -1 as expected (we're masking a difference)

Also have a look at the tests, they should help convince you as well that the logic is correct.

> Could it be that in the future you want to mask not just by position, but also depending on the letter in the sequence? In this case perhaps a list of concrete mutations to ignore would suite better. But that would not be practical if you want to ignore large ranges. Or we can have both somehow.

We usually mask ranges when making trees. A possible extension could be to mask different ranges for different purposes: e.g. don't count certain mutations towards private mutations/reversions in QC. The default could be to not penalize mutations in these regions if labelled/reverted, simply count as private mutations - but that may be overcomplicating things.

> Nextclade is not SC2-only software, so we might want to think about how it would serve for the analysis of other pathogens as well.

Definitely, masking terminals is very common across all pathogens - it's just that for SC2 our reference tree is most precise so I notice these things more. Richard may have more thoughts on generalisability/limitations. I made sure to allow specification of arbitrary ranges, not just terminals. It may be enough in most cases to just do terminals, but it's cheap to extend to arbitrary masked regions (`Vec<Range>` instead of `Option<Range>`)

ivan-aksamentov:
> If you trust the distance calculation without mask, and trust the code changing the query substitutions (and missing) then you should trust the new function - that's the beauty of the way it's implemented.

What I want to understand is whether adding a penalty to the distance this way for all sequences (good and bad) makes bad sequences less bad compared to good sequences. That is, by adding a constant linear bias to the formula, does it decrease the penalty for Ns in bad sequences?

See an example: Assume all other scores are 0. Before this PR: we have a "good" sequence with no Ns, which has `undetermined_sites` score of 0, and a "bad" sequence with 300 Ns and `undetermined_sites` 300. Now in this PR you mask 300 nucs. The "good" sequence now has `undetermined_sites` of 300 and the "bad" one, depending on where the masked range is, might have `undetermined_sites` anywhere from 300 to 600. That is, they now can potentially have the equal distance. 

Even if it can happen in terms of calculation, it might still not be relevant in terms of bioinformatics, but I felt like I need to highlight this possible case as an engineer.

corneliusroemer:
I'm not sure why you mention `undetermined_sites` - that is just an internal variable that is linked to the particular implementation (set overlap) but not intrinsically of interest.

What matters is distance. Distance is always bigger or equal to 0, and the effect of masked ranges is to make that distance smaller - but not bigger.

Yes, we throw away some information this way, masked sites are not used to resolve placement, but that's on purpose as these masked sites could sometimes _mislead_ and do harm, breaking a tie in the wrong direction, just due to noise.

Since we now break ties using priors, but only in case of equal distance, breaking ties due to noise does harm.

In case of SC2, we only mask 1%, the noisiest part, making results more interpretable and deterministic.

ivan-aksamentov:
@corneliusroemer Could you please pull updates from master, which contains a fix for the CI, to see how whether the web app build works?

ivan-aksamentov (approved):

corneliusroemer:
> Could you please pull updates from master, which contains a fix for the CI, to see how whether the web app build works?

Done!

corneliusroemer:
I'm happy to document the calculation, it's just a bit of set intersection stuff, like Venn diagrams.

rneher:
LGTM -- github having issues, wouldn't let me add a review to approve.

PR #1131, opened 2023-03-21
fix(web): avoid double analysis launch when using `input-fasta` param

ivan-aksamentov:


corneliusroemer:
Hmm, still getting the same double run when testing with: https://nextclade-git-fix-web-input-fasta-dupes-nextstrain.vercel.app/?dataset-name=sars-cov-2&input-fasta=example

ivan-aksamentov:
Yep, I tried a few things, but it does not work. Need to think about other solutions.

PR #1132, opened 2023-03-21
feat(cli): change gene warnings severity to info

ivan-aksamentov:
This changes gene warnings severity to info. As discussed on Slack, there are many cases where missing genes is not a problem, notably in case of partial sequences. This is to declutter the logs - warning is the default severity. Info is opt-in.


corneliusroemer (approved):
This is great - we should also remove the warnings from the web UI for missing genes

This is not done in this PR, correct? Yes it isn't: 
image

PR #1137, opened 2023-04-03
Update website copy

corneliusroemer:


ivan-aksamentov:
@corneliusroemer I just learned that [WHO recommends](https://www.who.int/news/item/28-11-2022-who-recommends-new-name-for-monkeypox-disease) Mpox as the new name for Monkeypox. But could we keep both names for the time being? This text mostly exists for mild [keyword stuffing](https://en.wikipedia.org/wiki/Keyword_stuffing) of search engines, such as Google, so that people could find Nextclade easier.

We have

```
Influenza (Flu)
Respiratory Syncytial Virus (RSV)
```

so perhaps we can say either of

```
Mpox (Monkeypox)
Monkeypox (Mpox)
```
?

We could also add some other keywords, such as flu flavors, to optimize relevance.

PR #1144, opened 2023-05-02
Try suggestion from chatGPT to accept empty string as valid result

corneliusroemer:
Hmm, this attempt doesn't work


ivan-aksamentov:
ChatGPT did not understand the code. It wrote the same thing, but worse. Let's try the good old IvanGPT this time.

Issue #1139, opened 2023-04-26
BUG: User's default dataset not getting version update unless they click "change"


corneliusroemer:
TL;DR: If you're a Nextclade user and seem to not get a new dataset version, please click "Change" and reselect that datset. That should fix it (until we've patched the bug).

We've had at least 3 reports of users getting old datasets after new datasets were released.

Based on user feedback, my assumption was thus far that users were using the same session, not reloading or restarting their machines. However, in retrospect this is quite unlikely given that one user reported having a dataset that was >2 months out of date (they will have most likely restarted their machine in between).

@emmahodcroft managed to catch the bug (getting old dataset) and we recorded what happened. In her case, opening a new tab she was shown an old dataset version (March instead of April)

Turns out that the index_v2.json _is_ loaded correctly (not cached), but it seems that we store the user's selected dataset across tabs. And that dataset is **pinned** to a particular **dataset version** not just **dataset type**.

A workaround for users is to simply click the "change" button which will unset the preference variables and pull in new version information from the index.

I never encountered this bug in my testing because I change datasets all the time. It's hard to reproduce for a frequent user. Thanks to all the users who reported: @JosetteSchoenma, Pavitra Roychoudhury and Emily Smith

Screenshot from Emma opening Nextclade in fresh tab:
<img width="1143" alt="image" src="https://user-images.githubusercontent.com/25161793/234619476-5395861d-4a3d-48af-acc7-cc2c2db56a59.png">

Emma has a detailed browser session recording available if required for debugging @ivan-aksamentov

PR #1147, opened 2023-05-02
Avoid stale datasets in Nextclade Web

ivan-aksamentov:
Resolves https://github.com/nextstrain/nextclade/issues/1139

This attempts to resolve situations described in the linked issues, when the new datasets are not available for users with stale browser tabs and also unless they switch the current dataset periodically.

To mitigate this, I add 2 kinds of data updates:

 - Re-fetch dataset index and silently update the list of pathogens
 - Check current dataset against the local list of datasets, and notify user when there is an update

Both updates are happening every time when:
 - User visits for the first time
 - The new tab is opened
 - The browser tab with Nextclade Web is focused
 - User navigates between pages

Additionally, these updates also happen periodically:
 - index re-fetch every 12 hours (there is a network request for index file)
 - current dataset check every 5 seconds (there is no network request involved)

If an update found for the currently selected dataset, Nextclade shows a notification. There users can consult datasets changelog and apply the update by clicking the "Update" button.

### Demo

Notice when I click to the field, the dataset version changes:

(there is only one dataset in the list shown in this example, but in real application all datasets will be updated according to the new index JSON file)

https://user-images.githubusercontent.com/9403403/235728933-fd539e26-3b3f-44d4-98ab-8812ce3db212.mp4

Notice when I click on the page, the notification appears, and when I click to the "Update" button, the dataset version changes:

https://user-images.githubusercontent.com/9403403/235729166-705e4d8c-5966-4aa4-a462-775d28d8932b.mp4

### Future work
 - Instead of silent update of the index we might also add notifications for when the index is updated.
 - The updates are currently irreversible. We might allow users to select older dataset versions from the list.

### DIFF
--- a/packages_rs/nextclade-web/src/components/Main/DatasetCurrent.tsx
+++ b/packages_rs/nextclade-web/src/components/Main/DatasetCurrent.tsx
@@ -1,10 +1,10 @@
+import { isNil } from 'lodash'
 import React, { useCallback, useState } from 'react'
-
-import { Button, Col, Collapse, Row } from 'reactstrap'
-import { useRecoilValue, useResetRecoilState } from 'recoil'
+import { Button, Col, Collapse, Row, UncontrolledAlert } from 'reactstrap'
+import { useRecoilValue, useResetRecoilState, useSetRecoilState } from 'recoil'
 import styled from 'styled-components'
-
-import { datasetCurrentAtom } from 'src/state/dataset.state'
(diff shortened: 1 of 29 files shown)


ivan-aksamentov:
@corneliusroemer yep, would be great if you before the next data update you'd keep a few stale tabs in various browsers, with and without datasets selected. And then updated master branch in the data repo. Then the new features should manifest in this preview version. This should be a good-enough real-world test.

Or we can also merge the software to master and see if nextclade master picks up the same thing.

Issue #1143, opened 2023-05-02
BUG(web/table): Empty strings make sort misbehave, not sort properly


corneliusroemer:
When there are empty strings (no values) in a column one sorts by, the sort misbehaves:
- rows with empty strings are not sorted consistently (to either top or bottom)
- on large tables the sort globally misbehaves in that one can get multiple locally sorted chunks

Reproduce:
1. Open nextclade.org
2. Select sars-cov-2 dataset, load example dataset
3. Start analysis
4. Sort by WHO name
5. Observe results

<img width="960" alt="image" src="https://user-images.githubusercontent.com/25161793/235626777-71b7b847-4c28-43dd-8ddf-9e7d4e0d6c12.png">

The example above does not cause the only-locally sorted issue. For that, one, try this:
https://clades.nextstrain.org/?dataset-name=sars-cov-2-21L&input-fasta=https%3A%2F%2Flapis.cov-spectrum.org%2Fopen%2Fv1%2Fsample%2Ffasta%3ForderBy%3Drandom%26limit%3D1000%26country%3DGermany%26dateFrom%3D2021-12-15%26dateTo%3D2022-12-15%26host%3DHomo%2Bsapiens%26downloadAsFile%3Dtrue
and sort by "unaliased lineage"

<img width="812" alt="image" src="https://user-images.githubusercontent.com/25161793/235629115-c0217f88-2de8-4eb4-b62b-e220cc6e4928.png">

I suspect the issue is that empty strings are treated as `None` and not as `""` in some JS comparison function, or something like that.

ivan-aksamentov:
@corneliusroemer It seems that these values are `undefined`, not empty strings.

Nextclade Web uses [stable sort](https://en.wikipedia.org/wiki/Sorting_algorithm#Stability) (order of equivalent entries is preserved) and empty (`undefined`) values are not sorted, because we don't know how (JS might have some weird behavior from when it was slapped together hastily in 1992, but the correct computer-sciency way to say is that empty values are not comparable to strings). I tried to model the behavior for the well-known tabular software, such as Excel.

Consider this:
 - you have fresh results from the Lapis link (perhaps fewer sequences, because 1000 is a bit too much to wait)
 - you can sort by "Clade" column and all "outgroup" items will go to the top or bottom
 - note that numbers in "i" column for "outgroup" entries have consecutive values - stable sort ensures that the order of values that all have equal "outgroup" clade is preserved the way it was before sort
[...]

PR #1152, opened 2023-05-04
fix(web): sort empty values

ivan-aksamentov:
Resolves https://github.com/nextstrain/nextclade/issues/1143

The `undefined` values are not comparable to strings, so Cornelius felt that in presence of empty values in custom clade-like columns the table sort misbehaves.

Here I treat the `undefined` values in custom clade-like columns and for phenotype columns as empty strings before sorting them. This way they can be sorted lexicographically along with other string values.

This may or may not what we want, but let's give it a try.


corneliusroemer:
Amazing, thanks! Fixes the bug - not sure what you mean by "This may or may not what we want, but let's give it a try."

Before this PR, empty values messed up sort globally, now sort behaves well, that's what we want :)

corneliusroemer (approved):

PR #1172, opened 2023-06-15
fix: adapt the gap-open-close calculation to CDS segments

rneher:
the previous implementation just looped over genes. Now we loop over all cds segments for all cds of all genes.


rneher:
@ivan-aksamentov 

this iteration needs to go from `end -> begin` if the segment is on the negative strand (wrong on master as well, I think).

PR #1193, opened 2023-07-05
fix: exclude gaps and N from divergence calculation

rneher:


rneher:
We might want to add a method to `Nuc` that `is_ACGT` or similar.

ivan-aksamentov:
> We might want to add a method to Nuc that is_ACGT or similar.

@rneher There is https://github.com/nextstrain/nextclade/blob/182fddfb276bff0602cfb7d02a955094457ea2d7/packages_rs/nextclade/src/alphabet/nuc.rs#L40-L68

packages_rs/nextclade/src/analyze/divergence.rs
```diff
@@ -1,14 +1,18 @@
...
+  private_mutations: &[NucSub],
   divergence_units: &DivergenceUnits,
   ref_seq_len: usize,
 ) -> f64 {
   // Divergence is just number of substitutions compared to the parent node
-  let mut this_div = number_private_mutations as f64;
-
+  let mut this_div = private_mutations
+    .iter()
+    .filter(|m| !m.ref_nuc.is_gap() && m.qry_nuc.is_acgt())
```
ivan-aksamentov, line 14: I replaced manual comparisons with methods on `Nuc`

ivan-aksamentov:
Although this PR might be fixing a particular observed issue with divergence calculations, I think there are serious consistency issues on the base branch that should be addressed in a more general way.

The array named `private_nuc_mutations.private_substitutions` should not contain deletions (especially that it's a sibling to `private_nuc_mutations.private_deletions`). And the gaps in `ref_nuc` is what? Insertions? This is simply nonsense and I think we should find the root cause and fixed it at (or close to) the place where these arrays are created, and not by filtering them ad-hoc to only mitigate a particular effect (on divergence calculation in this case).

rneher:
thanks. The tree structure and its fields comes from the auspice tree, I believe. And that tree has gaps and the reverse as individual mutations on the tree.

rneher:
We previously only had a notion of private mutations on terminal branches of nodes that get attached to the tree. now we need look specifically at mutations on branches in the reference tree. both terminal and internal branches. We could clean this up in the pre-processing of the tree.

ivan-aksamentov:
Okay, so it means that the data structures on the base branch are chosen incorrectly. Behind the walls of loops I cannot see immediately how this can be addressed more generally. I'll merge with a "FIXME" comment and let's see if I manage to solve this puzzle on the base branch after I am done with reviewing and cleaning it up.

PR #1196, opened 2023-07-12
Feat/greedy tree builder 2 refactor get closest neighbor recursively rn

rneher:


rneher:
We might consider turning `calculate_divergence` into `calculate_branchlength`.  we then could avoid looking up parent divergence here: 

https://github.com/nextstrain/nextclade/blob/feat/greedy-tree-builder-2-refactor-get_closest_neighbor_recursively-rn/packages_rs/nextclade/src/tree/tree_builder.rs#L236

and calculate divergence as `target_node_div - branchlength`. 

If we change `attach_node` such that it takes `divergence` of the new node as an argument, we wouldn't need to look up as much and wouldn't need `divergence_unit` and `ref_seq_len` as arguments.

PR #1197, opened 2023-07-12
fix: ensure tree output is deterministic

ivan-aksamentov:
Currently due to thread scheduling uncertainty, analysis outputs can be stored in unknown order. This affects tree construction in unexpected ways.

Here I ensure that outputs are sorted according to the index of the corresponding input fasta entry before they enter tree construction. This has performance penalty, but ensures deterministic outputs.


ivan-aksamentov:
Merged into branch `v3` and we will release it in Nextclade v3.

PR #1199, opened 2023-07-13
feat: improved stripe construction and visualization

corneliusroemer:
PR merges into feat/better-seed-matching, not master or v3!


rneher:
I think we can shelf this for now.

PR #1203, opened 2023-07-19
fix: avoid dependency on sys libs for xz2 and bzip2 crates

ivan-aksamentov:
This adds "static" feature for xz2 and bzip2 crates to ensure we don't depend on any system libraries.

### DIFF
--- a/packages_rs/nextclade/Cargo.toml
+++ b/packages_rs/nextclade/Cargo.toml
@@ -63,8 +63,8 @@ zip = { version = "=0.6.6", default-features = false, features = ["aes-crypto", 
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 atty = "=0.2.14"
-bzip2 = "=0.4.4"
-xz2 = "=0.1.7"
+bzip2 = { version = "=0.4.4", features = ["static"] }
+xz2 = { version = "=0.1.7", features = ["static"] }
 zstd = { version = "=0.12.3", features = ["zstdmt"] }
 
 [dev-dependencies]


tsibley:
I can confirm that this removes a dynamic runtime dependency on liblzma:

```diff
 $ ldd ./target/debug/nextclade 
         linux-vdso.so.1 (0xbadc0ffee0ddf00d)
-        liblzma.so.5 => /lib/x86_64-linux-gnu/liblzma.so.5 (0xbadc0ffee0ddf00d)
         libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0xbadc0ffee0ddf00d)
         librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0xbadc0ffee0ddf00d)
         libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0xbadc0ffee0ddf00d)
         libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0xbadc0ffee0ddf00d)
         libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0xbadc0ffee0ddf00d)
         libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0xbadc0ffee0ddf00d)
         /lib64/ld-linux-x86-64.so.2 (0xbadc0ffee0ddf00d)
```

ivan-aksamentov:
@tsibley Great! I'll merge then.

Do you see if any of the remaining runtime deps are avoidable?

P.S. I tried to lookup what is "linux-vdso.so" and entered it as is into browser URL bar. Instead of searching it navigated to https://linux-vdso.so/. Same with https://libm.so/. Look like they are really into Linux in Somalia :laughing:

tsibley:
> Do you see if any of the remaining runtime deps are avoidable?

Nope, at least for a target that's not fully static MUSL. vdso is for syscall optimizations and is used automatically and internally. ld is the dynamic linking loader, also automatic and necessary. The rest are part of libc.

PR #1205, opened 2023-07-19
feat: add Newick output

ivan-aksamentov:
This adds ability to output trees in nwk (Newick) format.


rneher:
LGTM, but the text in the download dialog is maybe not optional

"Phylogenetic tree with sequences placed onto it, in Newick format. The tree is in Nextstrain format. Can be viewed for example in [icytree.org](https://icytree.org/)."

Maybe better "Phylogenetic tree with sequences placed onto it, in Newick format. Can be viewed in most tree viewers, including [auspice.us](https://auspice.us) or [icytree.org](https://icytree.org/)."

PR #1214, opened 2023-07-22
Simpler chop

ivan-aksamentov:
Resubmit of https://github.com/nextstrain/nextclade/pull/1212

### DIFF
--- a/packages_rs/nextclade/src/align/seed_alignment.rs
+++ b/packages_rs/nextclade/src/align/seed_alignment.rs
@@ -284,7 +284,7 @@ pub fn create_stripes(
     look_back_length = max(look_back_length, current_seed.offset - current_broad_seed.left_offset);
   }
   // terminate previous trapezoid where the new one will start and push
-  current_broad_seed.ref_end = (current_seed_end as isize - look_back_length) as usize;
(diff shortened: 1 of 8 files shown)


packages_rs/nextclade/src/align/seed_alignment.rs
```diff
@@ -143,186 +143,190 @@ pub fn seed_alignment(
...
-      params.excess_bandwidth,
-      params.max_indel,
-      params.allowed_mismatches,
+      qry_len as isize,
+      ref_len as isize,
+      params.terminal_bandwidth as isize,
+      params.excess_bandwidth as isize,
+      params.max_indel as isize,
+      params.allowed_mismatches as isize,
     )
```
ivan-aksamentov, line 165: This probably comes from the base branch, but I only noticed it here. There's many arguments of the same type - easy to accidentally confuse their order. Unfortunately, Rust does not have named params, but we could consider:

 - to pass `params` directly, without referencing 4 of its fields before the call
 - if these fields and other args of the same type need to be passed separately, then perhaps it makes sense to make a struct (e.g. `CreateStripesParams`) and aggregate all or some of the functions' args there

ivan-aksamentov:
I accidentally merged incorrectly outside of GitHub. Now resubmitted as https://github.com/nextstrain/nextclade/pull/1220

PR #1217, opened 2023-07-24
feat: use another impl of MessagePack for communication between langs

ivan-aksamentov:


ivan-aksamentov:
Turns out to be slower than passing simple JSON string https://github.com/nextstrain/nextclade/pull/1215

PR #1219, opened 2023-07-24
Feat/validation workflow

corneliusroemer:
Add snakemake workflow that compares the current HEAD of nextclade vs a chosen branch, e.g. master, and produces various diffs of the output tsvs.

By default, this runs on all datasets using the example sequences, but it can also be run on different sequences, e.g. 20k SARS-CoV-2 sequences etc.


rneher:
@ivan-aksamentov This adds @corneliusroemer validation workflow (sort of similar to the smoke tests, but comparing a branch and the local copy) that produces nice diffs of the tsv outputs and the alignments. very helpful for spotting differences. Anything preventing this from being merged?

ivan-aksamentov:
I reverted a couple of commits which are unrelated to the main idea.

@corneliusroemer Please consult the team (or me) before you are altering cargo config. For a project-wide config file it is surprisingly wonky and brittle, and can lead to hard-to-find bugs. Especially it's a bad idea to snick them into unrelated PRs. Same for the dev docs.

You should not need to install `wasm-opt` manually - wasm-pack does this for you. If you think you need, then you probably hit [this](https://github.com/rustwasm/wasm-pack/issues/1263) bug in wasm-pack v0.11. Upgrading wasm-pack to the latest version should fix the issue.

Also, the [wasm-opt crate](https://crates.io/crates/wasm-opt) is some random Rust bindings to wasm-opt of unknown quality and it contains the binary you need only by accident. You probably don't want to keep this stuff on your machine or recommend it in the docs - it might be causing incompatibilities going forward. If you truly ever need a low-level tool like wasm-opt, then the only correct version is developed and shipped [here](https://github.com/WebAssembly/binaryen/releases).

I know that we are living in the high-dopamine, fast-moving world of ChatGPTs and of immediate StackOverflow solutions, but I won't be approving them into the repo my name is attached to.

ivan-aksamentov:
I don't entirely understand why this has to be Snakemake, but if it works for you, it works for me.

If you have a spare second, please add a requirements.txt or similar, so that we don't have to figure out dependencies at runtime.

Otherwise feel free to hit that "Merge" button.

ivan-aksamentov (approved):

ivan-aksamentov:
I tried to run it according to instructions, but it did not work for me. The error is not informative and logs don't seem to be  very useful:

 - [console.log](https://github.com/nextstrain/nextclade/files/12157044/console.log)

 - [run.log](https://github.com/nextstrain/nextclade/files/12157045/run.log)


Probably missing something, like deps. I am not a target audience of this, so if it works for you, then not a huge deal.

Also, and this is a matter of personal preference, but could we add `--quiet` flag to git and cargo commands to reduce the terminal lines racing to their life?

ivan-aksamentov:
Alright, that was my fault. I had `CARGO_BUILD_TARGET_DIR` set in my container environment quite irresponsibly.

I also added a few "fixes" that make it work for me. Feel free to revert of modify.

This is how I run it:

```bash
cd tests/diff-vs-base && pip3 install -r requirements.txt && snakemake
```

Now I can see a lot of very important greeny-yellowy snakemake things scrolling for a few kilometers and then a few directories appear. The problem I am facing is that I don't know what to do with that. I hypothesis that if there's no errors, then it's good to go? You might add a couple of sentences to the readme about what this whole thing does. But, again, other than you two, guys, no one else is probably going to be using that, so we might also keep it a secret knowledge. Or what is your plan for this functionality?

PR #1232, opened 2023-08-05
Feat/maximal alignment area

rneher:
This PR stubs out a check that the sum of the stripes (the area of the band) is smaller than a value to be provided by the command line (could be something 100_000_000 by default, each entry requires one `i32` and one `i8`). 

a commandline flag `--max-band-area` would replace `--max-indel` we used to have.


ivan-aksamentov:
Closing in favor of https://github.com/nextstrain/nextclade/pull/1233

PR #1233, opened 2023-08-09
feat: add max alignment band area check

ivan-aksamentov:
Resubmit of #1232, rebased on top of #1231 to avoid difficult merges.

Added `max_band_area` CLI and dataset parameter.

### DIFF
--- a/packages_rs/nextclade/src/align/params.rs
+++ b/packages_rs/nextclade/src/align/params.rs
@@ -47,6 +47,10 @@ pub struct AlignPairwiseParams {
   #[clap(long)]
   pub score_match: i32,
 
+  /// Maximum area of the alignment band. Alignments with large bands are slow to compute and require substantial memory in the current implementation. Alignment of sequences requiring band with area larger than this value, will not be attempted and a warning will be emitted.
+  #[clap(long)]
+  pub max_band_area: usize,
+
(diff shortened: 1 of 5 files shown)


packages_rs/nextclade/src/align/params.rs
```diff
@@ -47,6 +47,10 @@ pub struct AlignPairwiseParams {
   #[clap(long)]
   pub score_match: i32,
 
+  /// Maximum area of the alignment band. Alignments with large bands are slow to compute and require substantial memory in the current implementation. Alignment of sequences requiring band with area larger than this value, will not be attempted and a warning will be emitted.
+  #[clap(long)]
+  pub max_band_area: usize,
```
ivan-aksamentov, line 52: CLI help message is a reworded message from `max_indel`. To be improved.

packages_rs/nextclade/src/align/params.rs
```diff
@@ -143,7 +147,8 @@ impl Default for AlignPairwiseParams {
       penalty_gap_open_out_of_frame: 8,
       penalty_mismatch: 1,
       score_match: 3,
-      max_indel: 400,        // to be replaced
+      max_band_area: 500_000_000,
```
ivan-aksamentov, line 150: Default value is here. 

The usual precedence order apply:

 - CLI param (or Web param if we ever have it in the UI) has the highest priority
 - Then dataset defaults from `virus_properties.json`/`pathogen.json`
 - If none of that, than this hardcoded value

rneher (approved):

PR #1240, opened 2023-08-30
feat/better seed debug logging

corneliusroemer:
- feat: log basic seed and align stats at debug level

We're currently not really using `debug` level logging for anything. I think it'd be useful to output some basic stats for each sequence like what proportion of the query are covered by chained matches, mean band width, max indel, and things like that.

This is what this PR outputs at debug level (`-vv`):
```
2023-08-30 13:17:52.012 [D] compression.rs:68: When processing '"nextclade.tsv"': detected file extension 'tsv'. It will be using algorithm: 'None'
2023-08-30 13:17:52.012 [I] nextclade_loop.rs:87: Processing sequence 'Ireland/CO-NVRL-ecM21IRL00199074/2021'
2023-08-30 13:17:52.036 [D] seed_match2.rs:509: Chained seed stats. max indel: 12, # matches: 181, first/last match distance from start/end [start: (ref: 12, qry: 0), end: (ref: 129, qry: 0)], max unmatched stretch (ref: 173, qry stretch: 173)
2023-08-30 13:17:52.036 [D] seed_match2.rs:514: Seed alignment covers 87.09% of query length
2023-08-30 13:17:52.040 [D] align.rs:83: Nucleotide alignment band area=24370141, mean band width=814
2023-08-30 13:17:57.535 [D] align.rs:91: Attempt: 0, Alignment score: 86887
```

### DIFF
--- a/packages_rs/nextclade/src/align/score_matrix.rs
+++ b/packages_rs/nextclade/src/align/score_matrix.rs
@@ -38,7 +38,8 @@ pub fn score_matrix<T: Letter<T>>(
   let mut scores = Band2d::<i32>::new(stripes);
   let band_size = paths.data_len();
 
-  trace!("Score matrix: allocated alignment band of size={band_size}");
+  let mean_band_width = band_size / ref_len;
+  trace!("Score matrix: allocated alignment band of size={band_size}, mean band width={mean_band_width}");
 
   let left_align = match params.gap_alignment_side {
     GapAlignmentSide::Left => 1,
--- a/packages_rs/nextclade/src/align/seed_match2.rs
+++ b/packages_rs/nextclade/src/align/seed_match2.rs
@@ -12,6 +12,7 @@ use eyre::Report;
 use gcollections::ops::{Bounded, Intersection, IsEmpty, Union};
 use interval::interval_set::{IntervalSet, ToIntervalSet};
 use itertools::Itertools;
+use log::{debug, trace};
 use std::borrow::Cow;
 use std::cmp::{max, min};
 use std::collections::{BTreeMap, VecDeque};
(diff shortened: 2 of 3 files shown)


packages_rs/nextclade/src/align/score_matrix.rs
```diff
@@ -1,7 +1,7 @@
 use crate::align::band_2d::{Band2d, Stripe};
 use crate::align::params::{AlignPairwiseParams, GapAlignmentSide};
 use crate::alphabet::letter::Letter;
-use log::trace;
+use log::{debug, trace};
```
corneliusroemer, line 4: Unnecessary addition
```suggestion
use log::trace;
```

packages_rs/nextclade/src/align/seed_match2.rs
```diff
@@ -481,14 +482,44 @@ pub fn get_seed_matches2(
...
+      .unwrap_or(0);
+    let first_match = seed_matches.first().unwrap();
+    let last_match = seed_matches.last().unwrap();
+    let first_match_ref_pos = first_match.ref_pos;
+    let first_match_qry_pos = first_match.qry_pos;
+    let last_match_ref_pos = ref_seq.len() - last_match.ref_pos - last_match.length;
+    let last_match_qry_pos = qry_seq.len() - last_match.qry_pos - last_match.length;
+
+    let n_matches = seed_matches.len();
+    debug!("Chained seed stats. max indel: {max_shift}, # matches: {n_matches}, first/last match distance from start/end [start: (ref: {first_match_ref_pos}, qry: {first_match_qry_pos}), end: (ref: {last_match_ref_pos}, qry: {last_match_qry_pos})], max unmatched stretch (ref: {max_unmatched_ref_stretch}, qry stretch: {max_unmatched_qry_stretch})");
```
ivan-aksamentov, line 509: Could you please extract the body of the condition into a separate function (keeping the condition itself where it is), so that it does not distract from the main algo code.

Note that the conditional jump will always be executed in production. (And every log macro invocation also adds a similar conditional jump underneath, which might hurt performance in tight loops. This particular place seems to be executed rarely though.)


Is this thing required in release builds?

corneliusroemer: Good point to refactor into a function. 

I know that this will be always executed in production - but it's only executed once per sequence, I don't think we need to think about perf at things so rarely executed.

Not required in release builds but it's nice if once can always get this info - we also include lots of trace statements in release builds. What's the argument against including it?

PR #1241, opened 2023-08-30
fix: Use best alignment when max band size exceeded on retry

corneliusroemer:
Handle case when max band size is exceeded on retry. In this case, we already have a (potentially suboptimal) alignment that we can still use. Right now, we error unnecessarily by not handling this case (unwrapping stripe creation with `?`).

See https://neherlab.slack.com/archives/C015PFP5V44/p1693396367958919

I yet have to find a real life case where we hit this condition, will add a repro here if I find one

### DIFF
--- a/packages_rs/nextclade/src/align/align.rs
+++ b/packages_rs/nextclade/src/align/align.rs
@@ -66,8 +66,10 @@ pub fn align_nuc(
 
   let mut attempt = 0;
 
+  let mut pre_retry_alignment: Option<AlignmentOutput<Nuc>> = None;
+
   loop {
-    let stripes = create_alignment_band(
+    let stripes = match create_alignment_band(
       &seed_matches,
       qry_len as isize,
       ref_len as isize,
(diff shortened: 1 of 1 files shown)


packages_rs/nextclade/src/align/align.rs
```diff
@@ -66,16 +66,28 @@ pub fn align_nuc(
...
+    ) {
+      Ok(result) => result,
+      Err(err) => {
+        if let Some(alignment) = pre_retry_alignment {
+          info!("{err}");
+          info!("When processing sequence #{index} '{seq_name}': In nucleotide alignment on retry attempt {attempt}: Alignment area too large. Using previous alignment result. Alignment score was: {}", alignment.alignment_score);
+          return Ok(alignment);
+        }
+        return Err(err);
+      }
```
ivan-aksamentov, line 89: This assumes that the "Alignment area too large" is the only kind of error that can be emitted by the `create_alignment_band()`. Which is currently true, but is not guaranteed. If another error is returned, the message will be inaccurate.

If we return `band_area` from `create_alignment_band()`, the check can be performed outside, so that we know what happened in this case. Also it will avoid the punctuational storm and nested scopes.

packages_rs/nextclade/src/align/align.rs
```diff
@@ -66,16 +66,28 @@ pub fn align_nuc(
 
   let mut attempt = 0;
 
+  let mut pre_retry_alignment: Option<AlignmentOutput<Nuc>> = None;
```
ivan-aksamentov, line 69: The uninitialized variables is a code smell and a sign of either bugs or poorly structured code. 

On the other hand, I don't see immediately if this can be avoided with how the initial loop is written.

Does it mean we persist previous alignment result (including sequences), increasing memory consumption?

corneliusroemer: We don't really need to keep the alignment around once the stripes return successfully, so we could empty it again.

Does the alignment take much memory? It's not the stripes actual memory-costly scores/paths that are kept around here, only the result.

rneher:
closing this in favor of https://github.com/nextstrain/nextclade/pull/1247

PR #1243, opened 2023-09-01
fix: prevent run-away bandwith

rneher:


corneliusroemer (approved):
This significantly improves runaway bandwidth 
Bandwidth can still get large when `--excess-bandwidth` is on the order of ~200 or larger.
However, I'm not aware of datasets where we would need such large bandwidth.
This is in any case an improvement.
See https://neherlab.slack.com/archives/C015PFP5V44/p1693397022589159 for detailed investigation/testing

PR #1247, opened 2023-09-06
fix: return best alignment when band extension hits max_band_limit

rneher:
reimplementation of https://github.com/nextstrain/nextclade/pull/1241

this still complains about potential uninitialized variable.


rneher:
I think I have found a solution that avoids the uninitialized variable but does what we want it to do in a reasonably streamlined way.

ivan-aksamentov (approved):

ivan-aksamentov:
This is very nice! The code is so much prettier and easy to read now!

PR #1249, opened 2023-09-08
fix: zero length branches can still have aa mutations even without nu…

rneher:
…c muts

This problem arose when we are moving the attachment point across a branch of a node that had aa mutations that didn't correspond to any nucleotide mutations. The problem arose only in the specific case when the branch length of a terminal node is zero.

### DIFF
--- a/packages_rs/nextclade/src/tree/tree_builder.rs
+++ b/packages_rs/nextclade/src/tree/tree_builder.rs
@@ -193,6 +193,12 @@ pub fn finetune_nearest_node(
           current_best_node.payload().name
         )
       })?;
+      private_mutations = union_of_muts(&private_mutations, &best_split_result.left.invert()).wrap_err_with(|| {
+        format!(
+          "When calculating union of mutations between query sequence and the candidate child node '{}'",
+          graph.get_node(best_node.key()).expect("Node not found").payload().name
+        )
+      })?;
       current_best_node = graph
         .parent_of_by_key(best_node.key())
         .ok_or_else(|| make_internal_report!("Parent node is expected, but not found"))?;
--- a/packages_rs/nextclade/src/tree/tree_builder.rs
+++ b/packages_rs/nextclade/src/tree/tree_builder.rs
@@ -1,11 +1,11 @@
 use crate::analyze::aa_del::AaDel;
 use crate::analyze::aa_sub::AaSub;
-use crate::analyze::divergence::{calculate_branch_length, count_nuc_muts};
+use crate::analyze::divergence::{calculate_branch_length, count_nuc_muts, score_nuc_muts};
 use crate::analyze::find_private_nuc_mutations::BranchMutations;
 use crate::analyze::nuc_del::NucDel;
 use crate::analyze::nuc_sub::NucSub;
-use crate::graph::node::GraphNodeKey;
-use crate::make_internal_report;
+use crate::coord::range::NucRefGlobalRange;
+use crate::graph::node::{GraphNodeKey, Node};
 use crate::tree::params::TreeBuilderParams;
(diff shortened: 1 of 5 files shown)


packages_rs/nextclade/src/tree/tree_builder.rs
```diff
@@ -193,6 +193,12 @@ pub fn finetune_nearest_node(
...
       })?;
+      private_mutations = union_of_muts(&private_mutations, &best_split_result.left.invert()).wrap_err_with(|| {
+        format!(
+          "When calculating union of mutations between query sequence the zero-length parent node '{}'",
+          graph.get_node(best_node.key()).expect("Node not found").payload().name
+        )
+      })?;
       current_best_node = graph
         .parent_of_by_key(best_node.key())
         .ok_or_else(|| make_internal_report!("Parent node is expected, but not found"))?;
```
ivan-aksamentov, line 204: In terms of science:

 - Is the order of set operations significant here, and is it correct? I believe that `union(diff(pm, shared), left)` and `diff(union(pm, left), shared)` might yield different results.


A few technicalities:

 - The error message seems to be missing a word ("sequence and ..."?) or otherwise incomplete. It is also a bit asymmetric with the one from the diff operation.
 - I believe that the node name can be retrieved simpler, like in the diff operation. Unless it's a different name. But I think right now it just makes a round trip: node -> node key ->  (fallible operation) -> node -> payload -> name, instead of just node -> payload -> name.
 - Might need some more explanatory comments - the code flow in this function is so complex and poorly structured that I cannot understand a thing anymore, just a few weeks after its written. After a few bug fixes added, it will become gibberish even for you :) 


A few unrelated ideas for the future of this function (I am writing down to not forget):

 - I hope we can refactor control flow in this loop sometimes too.
 - In a single-threaded graph implementation perhaps generic `.payload()` makes less sense. Intrusive storage right in the node struct might make code more readable (shorter chains to access things).
 - The parallel variables: best_node, best_split_result, n_shared_muts - they are flowing together and are modified together. This is verbose and error-pron (can forget to modify one of them). Might be possible to aggregate them into one variable.
 - Lambda functions in error handling can be extracted into named functions to make the messages less intrusive.

ivan-aksamentov, line 204: In terms of science:

 - Is the order of set operations significant here, and is it correct? I believe that `union(diff(pm, shared), left)` and `diff(union(pm, left), shared)` might yield different results.


A few technicalities:

 - The error message seems to be missing a word ("sequence and ..."?) or otherwise incomplete. It is also a bit asymmetric with the one from the diff operation.
 - I believe that the node name can be retrieved simpler, like in the diff operation. Unless it's a different name. But I think right now it just makes a round trip: node -> node key ->  (fallible operation) -> node -> payload -> name, instead of just node -> payload -> name.
 - Might need some more explanatory comments - the code flow in this function is so complex and poorly structured that I cannot understand a thing anymore, just a few weeks after its written. After a few bug fixes added, it will become gibberish even for you :) 


A few unrelated ideas for the future of this function (I am writing down to not forget):

 - I hope we can refactor control flow in this loop sometimes too.
 - In a single-threaded graph implementation perhaps generic `.payload()` makes less sense. Intrusive storage right in the node struct might make code more readable (shorter chains to access things).
 - The parallel variables: best_node, best_split_result, n_shared_muts - they are flowing together and are modified together. This is verbose and error-pron (can forget to modify one of them). Might be possible to aggregate them into one variable.
 - Lambda functions in error handling can be extracted into named functions to make the messages less intrusive.

rneher:
regarding your concern on the order of union and difference. I don't think this is an issue because in `private_mutations` each position has to be unique. The same is true for the result of the split operation. A particular position can only be either in the `shared`, or in the `left` list. Since a particular position is touched at most in one of the two operations, the order of the operations doesn't matter.

PR #1250, opened 2023-09-11
Refactor and cleanup tree finetuning code

ivan-aksamentov:
Followup of https://github.com/nextstrain/nextclade/pull/1249

Inspired by refactoring in [`15757ec` (#1249)](https://github.com/nextstrain/nextclade/pull/1249/commits/15757ece18f4e57931c27306e18c2ae229b2cfdb) I decided to continue, because the code in the `finetune_nearest_node()` can be clearly divided into independent functions.

Change of terminology, to clarify meaning of varables:

 - `current_best_node` -> `best_node` (this is the loop's external state)
 - `best_node` -> `candidate_node` 
 - `best_split_result` ->  `candidate_split`

I tried to make commits to contain only 1 minimal change which compiles and runs. Plus there is a few pink elephants I encountered  - reported in the PR comments below.

### DIFF
--- a/packages_rs/nextclade/src/analyze/divergence.rs
+++ b/packages_rs/nextclade/src/analyze/divergence.rs
@@ -1,4 +1,6 @@
 use crate::analyze::nuc_sub::NucSub;
+use crate::coord::range::NucRefGlobalRange;
+use crate::tree::params::TreeBuilderParams;
 use crate::tree::tree::DivergenceUnits;
 
 /// Calculate number of nuc muts, only considering ACGT characters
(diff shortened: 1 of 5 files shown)


ivan-aksamentov:
"Is leaf and not root" should always be the case for trees with more than one node, probably redundant.

Old version:

https://github.com/nextstrain/nextclade/blob/15757ece18f4e57931c27306e18c2ae229b2cfdb/packages_rs/nextclade/src/tree/tree_builder.rs#L173-L175

I kept it as is in the new version:

https://github.com/nextstrain/nextclade/blob/a44299945d47ae77d9115059bbbbb40564509c1f/packages_rs/nextclade/src/tree/tree_builder.rs#L207

ivan-aksamentov:
Check for potential bug here.

This is a roundrip: node -> node key -> node, equivalent of no-op:

https://github.com/nextstrain/nextclade/blob/15757ece18f4e57931c27306e18c2ae229b2cfdb/packages_rs/nextclade/src/tree/tree_builder.rs#L169-L171

In the new version I replaced it with just the variable itself:

https://github.com/nextstrain/nextclade/blob/a44299945d47ae77d9115059bbbbb40564509c1f/packages_rs/nextclade/src/tree/tree_builder.rs#L205

ivan-aksamentov:
The control flow here is very convoluted:
 - there are repeated conditions
 - there are 5 points of exit (`return`/`break`)

Can this be simplified?
If not, could we add some comments for each condition and each point of exit?


Old version:

https://github.com/nextstrain/nextclade/blob/15757ece18f4e57931c27306e18c2ae229b2cfdb/packages_rs/nextclade/src/tree/tree_builder.rs#L159-L182


In the new version, I kept it mostly as is, but it looks simpler (due to removal of assignment to mutable vars):

https://github.com/nextstrain/nextclade/blob/a44299945d47ae77d9115059bbbbb40564509c1f/packages_rs/nextclade/src/tree/tree_builder.rs#L188-L213

ivan-aksamentov:
In general, refactoring involved extracting new scopes (functions) which breaks continuity of the original large scope. Because the code relied on modifying the global state a lot (assignment to mutable variables), I might have introduced bugs where the variables are not updated properly anymore. This needs to be double-checked.

PR #1252, opened 2023-09-11
Account for placement mask, compare nodes based on mut score

ivan-aksamentov:


packages_rs/nextclade/src/analyze/divergence.rs
```diff
@@ -24,3 +26,18 @@ pub fn calculate_branch_length(
...
+
+  // Split away masked mutations
+  let (muts, masked_muts): (Vec<_>, Vec<_>) =
+    nuc_muts.partition(|m| masked_ranges.iter().any(|range| range.contains(m.pos)));
+
+  let n_muts = muts.iter().count() as f64;
+  let n_masked_muts = masked_muts.iter().count() as f64;
+
+  return n_muts + n_masked_muts * params.masked_muts_weight;
+}
```
ivan-aksamentov, line 43: The score is computed here

packages_rs/nextclade/src/tree/params.rs
```diff
@@ -14,13 +14,17 @@ pub struct TreeBuilderParams {
...
 
 #[allow(clippy::derivable_impls)]
 impl Default for TreeBuilderParams {
   fn default() -> Self {
     Self {
       without_greedy_tree_builder: false,
+      masked_muts_weight: 0.5,
     }
   }
 }
```
ivan-aksamentov, line 30: The weight param and its default are defined here

packages_rs/nextclade/src/tree/tree_builder.rs
```diff
@@ -106,23 +107,26 @@ pub fn finetune_nearest_node(
...
           "When calculating shared mutations against the current best node '{}'",
           best_node.payload().name
         )
       })?;
 
     // Check if the new candidate node is better than the current best
-    let n_left_muts = count_nuc_muts(&candidate_split.left.nuc_muts);
-    match find_better_node_maybe(graph, best_node, candidate_node, n_shared_muts, n_left_muts) {
+    let left_muts_score = score_nuc_muts(&candidate_split.left.nuc_muts, masked_ranges, params);
+    match find_better_node_maybe(graph, best_node, candidate_node, shared_muts_score, left_muts_score) {
```
ivan-aksamentov, line 129: I use the new score for both "shared" and "left" muts, even though "left" is currenly only used to compare to 0 (i.e. it is equivalent to `.is_empty()`)

packages_rs/nextclade/src/tree/params.rs
corneliusroemer: Nice! I think we might want a lower default so that masks is only used to resolve otherwise identical séquences.

PR #1270, opened 2023-10-11
feat: deprecate old alignment params

ivan-aksamentov:
Here I deprecate some of the alignment params which lost their use after the alignment algo was changed.

I emit an error if any of these parameters are still present in CLI args and/or in dataset's `pathogen.json`. We could also make it a warning.


packages_rs/nextclade/src/align/params.rs
```diff
@@ -164,11 +163,49 @@ impl Default for AlignPairwiseParams {
...
+impl AlignPairwiseParams {
+  pub fn validate(&self) -> Result<(), Report> {
+    #[rustfmt::skip]
+  let deprecated = BTreeMap::from([
+    (o!("--min-seeds            (minSeeds)"),             &self.min_seeds),
+    (o!("--seed-length          (seedLength)"),           &self.seed_length),
+    (o!("--seed-spacing         (seedSpacing)"),          &self.seed_spacing),
+    (o!("--min-match-rate       (minMatchRate)"),         &self.min_match_rate),
+    (o!("--mismatches-allowed   (mismatchesAllowed)"),    &self.mismatches_allowed),
+    (o!("--max-indel            (maxIndel)"),             &self.max_indel),
```
ivan-aksamentov, line 192: These are the params being deprecated

packages_rs/nextclade/src/align/align.rs
```diff
@@ -46,7 +46,7 @@ pub fn align_nuc(
     );
   }
 
-  if ref_len + qry_len < (10 * params.seed_length) {
+  if ref_len + qry_len < (20 * params.kmer_length) {
     // for very short sequences, use full square
```
ivan-aksamentov, line 50: According to Richard's recommendation, for short sequence check let's use `kmer_length` instead of removed `seed_length` and adjust the coefficient to roughly match the old value.

Issue #700, opened 2022-01-25
ENH: Display ambiguous mutations in gene/nuc view tooltip


corneliusroemer:
Ambiguous nucleotides are currently displayed only in the tooltip of the corresponding column:
![image](https://user-images.githubusercontent.com/25161793/151008404-192877fc-f898-4406-96d8-d350ca9ec79d.png)

In gene view, ambiguous bases turn the corresponding AA into an X. It would be nice, if the X tooltip showed corresponding ambiguous nucleotides, like AA mutation tooltips show the mutated nucs.

Current:
![image](https://user-images.githubusercontent.com/25161793/151008857-822ef143-8792-4fba-9b87-252121a6c8b9.png)

Something like this would be cool, potentially even with a range of possible amino acids that is compatible with the ambiguity.
![image](https://user-images.githubusercontent.com/25161793/151008890-cde13239-1869-4bcf-9092-00e186565f9a.png)

ivan-aksamentov:
The translation of ambiguous nucs is a cool feature we discussed with Richard even in the early days of Nextclade. It can be implemented quite easily, by modifying the [decoding map](https://github.com/nextstrain/nextclade/blob/af35dfa8fed04c362ec80aac2650d0a45fdbed7d/packages_rs/nextclade/src/translate/translate.rs#L11-L78). Someone needs to actually build the map and put triplet-to-aa entries above the catch-all case which resolves to AA `X`.

The coolness of this I see in that it will reduce the number of `X`, and salvage some amount of useful info even from mediocre data.

Regarding showing how ambiguous nucs and AAs are linked - it's trickier. We currently don't track why an AA became an `X` and there are multiple reasons. Also X are tracked as ranges, and within a range there could be multiple reasons by itself, so some deeper code changes will be needed.

corneliusroemer:
Just wanted to open a duplicate then found this: I was surprised that we don't indicate ambiguous nucleotides in the nuc view. Quite unexpected, I think we don't need to overcomplicate things: just add them like a single missing nuc.
<img width="174" alt="image" src="https://github.com/nextstrain/nextclade/assets/25161793/2ed6f0bd-0290-4978-aec6-d20af7c8b329">

corneliusroemer:
Some shade of grey would be the obvious choice I think. Doesn't matter so much what it is, as long as there's something that can be hovered to see extra information.

[...]

PR #1273, opened 2023-10-18
feat(web): add sequence markers for ambiguous nucleotides

ivan-aksamentov:
Resolves: https://github.com/nextstrain/nextclade/issues/700

Add colored markers for ambiguous (non-ACTGN) nucleotides to nucleotide sequence view.

I tried to pick colors to be pale (so that they resemble pale grey of missing nucs) and to be a mixture of colors of corresponding disambiguated characters, or to be inverse color in case of "not" characters (e.g. color of not-A is a pale inverse of color of A).

I made markers half-height at the top by default, just like missing nucs. (Settings UI is currently broken, so it's impossible to change at the moment, sorry)

Tooltip text is as specified in the original issue.


corneliusroemer:
Thanks that's excellent!

corneliusroemer (approved):

PR #1296, opened 2023-10-30
feat: Separate undeletions from reversion substitutions

corneliusroemer:
In v2, we ignored undeletions (i.e. ref is gap, qry is ref)
Until now, in v3, we treat them as reversion substitutions
A single stretch 20 bps undeleted results in 20 reversion substitutions which results in unexpectedly large QC penalties (some penalty is ok but it's too large as is)

In this PR, I split undeletions from ordinary substitutions and then aggregate them into NucUndelRanges (Richard said you might have a function for that @ivan-aksamentov but I couldn't find one so I made one myself)

Currently, nothing is done with the undeletion ranges, but we could surface them into TSV/web UI at a later point. We could also count each stretch towards QC as a single unit of a reversion (and analogously treat each private deletion stretch as a single private mutation).

We also don't seem to surface private deletions at the moment, so this could be done as a package.

### DIFF
--- a/packages_rs/nextclade/src/analyze/find_private_nuc_mutations.rs
+++ b/packages_rs/nextclade/src/analyze/find_private_nuc_mutations.rs
@@ -5,6 +5,7 @@ use crate::analyze::is_sequenced::{is_nuc_non_acgtn, is_nuc_sequenced};
 use crate::analyze::letter_ranges::NucRange;
 use crate::analyze::nuc_del::{NucDel, NucDelRange};
 use crate::analyze::nuc_sub::{NucSub, NucSubLabeled};
+use crate::analyze::nuc_undel::{NucUndel, NucUndelRange};
 use crate::analyze::virus_properties::{NucLabelMap, VirusProperties};
 use crate::coord::position::{NucRefGlobalPosition, PositionLike};
 use crate::coord::range::NucRefGlobalRange;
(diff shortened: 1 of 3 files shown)


corneliusroemer:
The difference in behaviour seems to be due to a replacement of: `let is_not_node_deletion = !node_qry.is_gap();` in v2 with     `let pos_is_non_acgtn = is_nuc_non_acgtn(pos, non_acgtns);` in v3

These check different things - to undo the difference between v2 and v3 without any of the filtering of undeletions, we could just add the ` let is_not_node_deletion = !node_qry.is_gap();` check back in. @ivan-aksamentov do you know the reasoning behind removing it? I think the changes were made here: https://github.com/nextstrain/nextclade/pull/1224

Could it be that you thought that the `non_acgtn` test _includes_ gaps as `non_acgtn` but it tursn out it doesn't?

Here's v2: https://github.com/nextstrain/nextclade/blob/f4cb9340684cc87787b313fd470387669b4ff01d/packages_rs/nextclade/src/analyze/find_private_nuc_mutations.rs#L248C1-L248C1
Here's v3: https://github.com/nextstrain/nextclade/blob/f214f301fea32cde418d58360406cf70a1b7ee39/packages_rs/nextclade/src/analyze/find_private_nuc_mutations.rs#L279

corneliusroemer:
This PR doesn't seem to break tree building for undeleted sequences, which may have been a worry: 
image
The screenshot shows how two sequences with undeletions of type a) are treated by tree building. Since this PR only changes treatment of type a) undeletions, that's what needs to be checked.

corneliusroemer:
Closed in favor of #1301  for now (we might still want to do aggregation at a later point)

PR #1298, opened 2023-10-30
Consistent usage of gene vs CDS in the interface

ivan-aksamentov:
Attempts to remove inconsistencies in naming and handling of genes and CDSes, to reflect changes in how genome annotations are handled in v3.

In Nextclade v2 there were only genes, so in Nextclade CLI `--genes` argument and `{gene}` placeholder in `--output-translations` make sense.

In Nextclade v3 there are genes and CDSes involved and CDSes have central place - they are now what's need to be selected by the `--gene` arguments and the CDS name needs to be substituted where the `{gene}` variable was.

So in this PR, at the very minimum, I did:

 - rename `--genes` to `--cds-selection` (initially `--cdses`, but Richard suggested `--cds-selection`)
 - rename `{gene}` to `{cds}` in `--output-translations` template string
 - change default value for `--output-translations` template string from `nextclade_gene_{gene}.translation.fasta` to `nextclade_cds_{cds}.translation.fasta` (breaking, because there is a word "gene" changed to "cds". Not the variable, it disappears when substututed, but the hardcoded word - it stays).
 - change template string in Nextclade Web from `nextclade.peptide.{gene}.fasta` to `nextclade.peptide.{cds}.fasta` (transparent to users)

This is a breaking change and involves changes to CLI arguments and to default output filenames.

#### Open questions:
 - Whether referring to CDSes by name only is sufficient. In general case, in GFF3, there could be multiple CDS with the same name. They can be nested under different genes or not.
 - If it's not sufficient, then we might consider a different design, where both gene name and CDS name are involved. For example, `--cdses` can accept a hierarchical name in `{gene}/{cds}` format and `--output-translations` would allow both `{gene}` and `{cds}` placeholders. Both can be optional features - you can omit `{gene}`.
 - Alternative design is separate `--cdses` and `--genes` flag, but then linking of the two lists of values is impossible in order to resolve potential duplicate names.
 - The root of the problem is that the GFF3 format identifies features by index of the row, and we are trying to resolve features by name. And in case of `--output-translations`, filesystems require filenames to be unique . This is similar to the problem of resolving sequences by sequence names in fasta (solved by using index in the fasta file instead). This is a problem for downstream users as well, because in case of ambiguity of names, it's impossible to find the required output file.
 - If we want to take into account the order, we might add variables which denote index of gene and cds, e.g.  `{i_gene}`, `{i_cds}`. Indices to GFF3 rows are guaranteed to be unique. This would allow users to also tie output translations with annotation or with mutations.
 - Indices is how I initially wanted to keep track of GFF3 features internally, but unfortunately our parser library does not provide row indices of features. We can fork the code though.
 - Our mutation notation only contain CDS now: `{cds}:{ref}{pos}{qry}`. In case of non-unique CDS names it's impossible to tell which one the mutation belongs to. To solve this we could use hierarchical notation as well: `{gene}/{cds}:{ref}{pos}{qry}` and perhaps simplify the notation automatically if CDS and gene have the same name, e.g. avoid `S/S:`. Can also be made configurable on runtime. But this variable format will be fun to parse in downstream applications. I cannot see how indices can be used here - technically they could be used, but in practice this will be very hard to read.
 - CDS selector dropdown currently can list CDSes with non-unique names, but only if they are nested under genes with different names.
 - Difficulty of the problem increases, if/when we add proteins (`--proteins`? `{protein}`?)

### DIFF
--- a/packages_rs/nextclade-cli/src/cli/nextclade_cli.rs
+++ b/packages_rs/nextclade-cli/src/cli/nextclade_cli.rs
@@ -816,7 +816,7 @@ pub fn nextclade_get_output_filenames(run_args: &mut NextcladeRunArgs) -> Result
 
     if output_selection.contains(&NextcladeOutputSelection::Translations) {
       let output_translations_path =
-        default_output_file_path.with_file_name(format!("{output_basename}_gene_{{gene}}"));
+        default_output_file_path.with_file_name(format!("{output_basename}_cds_{{cds}}"));
       let output_translations_path = add_extension(output_translations_path, "translation.fasta");
 
       let output_translations_template = output_translations_path
--- a/packages_rs/nextclade-cli/src/cli/nextclade_cli.rs
+++ b/packages_rs/nextclade-cli/src/cli/nextclade_cli.rs
@@ -350,7 +350,7 @@ pub struct NextcladeRunInputArgs {
   /// not be translated, amino acid sequences will not be output, amino acid mutations will not be detected and nucleotide sequence
   /// alignment will not be informed by codon boundaries
   ///
-  /// List of genes can be restricted using `--genes` flag. Otherwise all genes found in the genome annotation will be used.
+  /// List of genes can be restricted using `--genes` flag. Otherwise, all genes found in the genome annotation will be used.
   ///
   /// Overrides genome annotation provided by the dataset (`--input-dataset` or `--dataset-name`).
   ///
@@ -362,11 +362,12 @@ pub struct NextcladeRunInputArgs {
   #[clap(value_hint = ValueHint::FilePath)]
   pub input_annotation: Option<PathBuf>,
 
-  /// Comma-separated list of names of genes to use.
+  /// Comma-separated list of names of coding sequences (CDSes) to use.
   ///
   /// This defines which peptides will be written into outputs, and which genes will be taken into account during
-  /// codon-aware alignment and aminoacid mutations detection. Must only contain gene names present in the genome annotation. If
-  /// this flag is not supplied or its value is an empty string, then all genes found in the genome annotation will be used.
+  /// codon-aware alignment and aminoacid mutations detection. Must only contain CDS names present in the genome annotation.
+  ///
+  /// If this flag is not supplied or its value is an empty string, then all CDSes found in the genome annotation will be used.
   ///
   /// Requires `--input-annotation` to be specified.
   #[clap(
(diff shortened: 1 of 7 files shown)


corneliusroemer:
Good change!

tsibley:
Initial thoughts upon a first read.

The general strategy I'd suggest is to make predictable and fully-resolved forms of id possible in input/output while still allowing some alternative forms for convenience as long as they resolve precisely.

For input (e.g. `--cds-selection` and similar), as a convenience, allow identifying CDS by name when there's only one CDS with that name. When there's more than one, require explicit disambiguation either by name _or_ by index, e.g. by `{gene}/{cds}` or by `#{i_gene}.{i_cds}`. But make sure that the separator (e.g. `/` or `.`) is unambiguous and can not be a valid part of either a gene or CDS name. Always allow the explicit precise input, even if unnecessary (e.g. when CDS name alone would be sufficient), for users or programmatic usage that want predictability (at the expense of verbosity). Allowing disambiguation by name or by index will suit different users/use cases. And it sounds like sometimes disambiguation by name will not be sufficient and index will be required?

For output (e.g. `--output-translations` and mutation notation), default to a precise unambiguous form but include additional, not-strictly-necessary information for human users. For example, `nextclade_cds_{i_gene}.{i_cds}_{cds}.translation.fasta` or similar. Allow ambiguous forms, but document they may cause issues. Or better yet, if possible, detect when they will or will not cause issues and error/refuse to run if they will (e.g. require that for `--output-translations nextclade_cds_{cds}.translation.fasta` to be allowed, all CDS given to `--cds-selection` must have unique names). The same approach can be used with mutations, e.g. `#{i_gene}.{i_cds}:{gene}/{cds}:{ref}{pos}{qry}`.

> * rename `--genes` to `--cds-selection` (initially `--cdses`, but Richard suggested `--cds-selection`)

FWIW, I think `--cds` would be just fine, even if it's not unambiguously plural.

rneher:
I agree that just using `--cds` is fine. I just didn't like the 'forced plural' `--cdses`. 

I am not sure using hierarchical naming schemes involving gene and cds will completely solve this problem. I think we should either (i) insist on unique CDS names, or (ii) augment them with the ID or index of the feature if necessary. I am leaning towards requiring them to be unique. 

Why don't we change the default for `--output-translations` and in web to `nextclade.cds_translation.{cds}.fasta`. Then we have flexibility to add proteins later and remove the `peptide` which is a bit unusual.

ivan-aksamentov:
Alright, I have changed the default filenames according to what Richard suggested, uniformly for CLI and Web.

And I will enforce unique CDS names in a separate PR, as a part of GFF validation. This is although a crude solution for the identification problem, but is very simple to do, and will get us going faster for now.

Then, if/when we decide that we need a more sophisticated approach, we'll break the interface in a major version update.

PR #1303, opened 2023-10-31
fix: correctly filter CDSes in genome annotation

ivan-aksamentov:
Genome annotation was not filtered correctly. The values from `--cds-selection` (previously `--genes`) were still used to filter genes, not CDSes.

Here I reimplement filtering, such that:

- first CDSes are filtered for each gene
- then empty genes are dropped

At the same time, I move gene storage from a map to vec, to prepare it for the better genome annotation validation (this allows me to avoid reimplementing the filtering twice - for map and then for vec, even though it's unrelated to this particular PR).


ivan-aksamentov:
The errors in smoke tests will be addressed in https://github.com/nextstrain/nextclade/pull/1304

PR #1305, opened 2023-11-02
fix(web): spawn more than one analysis worker

ivan-aksamentov:


corneliusroemer:
We spawned only one analysis worker instead of `numThreads`, meaning we didn't use any multiprocessing for analysis. This is now fixed and we're making much better use of multicore CPUs. The effective speedup is on the order of a factor of 2 in a few brief benchmarks for both Ivan and me.

PR #1313, opened 2023-11-13
feat(web): default to official repo and path when omitted from url param

ivan-aksamentov:
This allows to use GitHub shortcuts with only branch specified:

```
?dataset-server=gh:@my-branch@
```

In this case, it assumes official data repo (`nextstrain/nextclade_data`) and the default path (`/data_output`) in it.


corneliusroemer:
That's neat!

corneliusroemer (approved):
Tested with https://nextclade-git-feat-web-url-params-optional-nextstrain.vercel.app/?dataset-server=gh:@mpxv-update-draft@ and it works!

ivan-aksamentov:
This unffortunately introduces a few bugs, notably the `dataset-server=gh` (without colon and without value) throws an error, but this codepath is important for Vercel to function. I need to enumerate all the URL params and formats to see what we need to support exactly, and refactor this part of the code - it's a bit convoluted currently and this leads to some corner cases not being handled properly.

PR #1329, opened 2023-12-07
Rename defaultGene and geneOrderPreference

ivan-aksamentov:
Sibling PR in data: 

`defaultGene` -> `defaultCds`
`geneOrderPreference` -> `cdsOrderPreference`

Data need to be adjusted first and data PR needs to be merged simultaneously with this PR.


corneliusroemer:
Change for SC2 was a one line change, mpox didn't have to be renamed as it doesn't use these two keys (yet)

PR #1348, opened 2023-12-18
feat(web): Support Safari >= 16.0 again

corneliusroemer:
The reason we deprecated Safari was that it didn't support shared web workers Since v16.0 shared web workers are supported so we no longer need to show the browser warning  Alternatively, we could discourage less loudly, saying that Safari is not tested (for >=16) while still erroring for lower versions.


ivan-aksamentov:
@corneliusroemer There has been problems beyond web workers, for example with CSS. Do the `/results`  and `/tree` pages look reasonable on Safari >= 16?

corneliusroemer:
Everything looks good on Safari 17.2 - I can keep using it there for a bit to see how robust it is.

image

image

PR #1377, opened 2024-01-10
feat: set gap alignment to "left" by default

ivan-aksamentov:


corneliusroemer:
Not blocking but it would be nice if we could indicate the default value in the CLI help. Right now it says "possible values left, right" but not which one is the default.

ivan-aksamentov:
@corneliusroemer I made it as follows now:

```
      --gap-alignment-side <GAP_ALIGNMENT_SIDE>
          Whether to align gaps on the left or right side if equally parsimonious. Default: left
          
          [possible values: left, right]

```

ivan-aksamentov:
This text could be incorrect though depending on how dataset is configured, which we have no way of telling at the time of printing help text.

corneliusroemer:
Thanks, that's great! I think it's ok that pathogen json can overrides default and CLI won't show. The CLI just shows the bare default that applies in case nothing overrides.

ivan-aksamentov:
@corneliusroemer We have a failing test now. Can you help me with it?
https://github.com/nextstrain/nextclade/actions/runs/7477824073/job/20351461502?pr=1377#step:7:397

rneher:
The validating the path matrices is a bit tricky, but it all looks correct to me.

Issue #907, opened 2022-06-29
Last component of output base names containing multiple periods is dropped


KatSteinke:
I've run into an issue when specifying an `output-basename` that contains multiple periods such as e.g. `SAMPLENAME.very.long.basename` in Nextclade 2.0.0 - the last component of the name is consistently dropped, so result files are named e.g. `SAMPLENAME.very.long.[result].[extension]` . 
For more detail: I'm running this with `output-all`, `output-basename`, `output-translations` and `output-csv` specified; the latter two are absolute paths, and work without problems.

corneliusroemer:
Thanks for reporting. I can reproduce this:
```
nextclade dataset get --name='sars-cov-2' --output-dir=data
nextclade run -D data --output-all out --output-basename basename.THISWILLDISAPPEAR -t out/test.tsv data/sequences.fasta --output-selection fasta
ls out
```

The result is:
```txt
basename.aligned.fasta  test.tsv
```

As you report, `.THISWILLDISAPPEAR` has disappeared.

corneliusroemer:
Root cause:
https://github.com/nextstrain/nextclade/blob/e9df843c4232d333cca3c6bebd7e14895a2abf3d/packages_rs/nextclade-cli/src/cli/nextclade_cli.rs#L631

`with_extension()` _replaces_ the extension if present. There's no stdlib way to `add_extension` yet unfortunately.

ivan-aksamentov:
This was fixed in https://github.com/nextstrain/nextclade/pull/913 and will be [released](https://github.com/nextstrain/nextclade/releases) in 2.0.1 [in a few moments](https://app.circleci.com/pipelines/github/nextstrain/nextclade/3307/workflows/0081055b-b59d-4601-a506-2e7938444e98).

PR #1387, opened 2024-01-22
fix: remove extra "translations.fasta" in default output translation path

corneliusroemer:
It seems that this line should have been removed in https://github.com/nextstrain/nextclade/pull/1298
but this was forgotten.

The default path will now be `nextclade.cds_translation.{cds}.fasta` as Richard suggested in
https://github.com/nextstrain/nextclade/pull/1298#issuecomment-1786602733

Before this PR it was (wrongly) `nextclade.cds_translation.{cds}.fasta.translation.fasta` since v3 release.

The default output path rename is a v3 breaking change that we don't seem to have documented yet. How shall we best do this given that v3.0.0 is already out? Post hoc editing changelog and adding to migration docs? Or do you prefer something else @ivan-aksamentov?

### DIFF
--- a/packages_rs/nextclade-cli/src/cli/nextclade_cli.rs
+++ b/packages_rs/nextclade-cli/src/cli/nextclade_cli.rs
@@ -835,7 +835,6 @@ pub fn nextclade_get_output_filenames(run_args: &mut NextcladeRunArgs) -> Result
     if output_selection.contains(&NextcladeOutputSelection::Translations) {
       let output_translations_path =
         default_output_file_path.with_file_name(format!("{output_basename}.cds_translation.{{cds}}.fasta"));
-      let output_translations_path = add_extension(output_translations_path, "translation.fasta");
 
       let output_translations_template = output_translations_path
         .to_str()
--- a/packages_rs/nextclade-cli/src/cli/nextclade_cli.rs
+++ b/packages_rs/nextclade-cli/src/cli/nextclade_cli.rs
@@ -834,7 +834,8 @@ pub fn nextclade_get_output_filenames(run_args: &mut NextcladeRunArgs) -> Result
 
     if output_selection.contains(&NextcladeOutputSelection::Translations) {
       let output_translations_path =
-        default_output_file_path.with_file_name(format!("{output_basename}.cds_translation.{{cds}}.fasta"));
+        default_output_file_path.with_file_name(format!("{output_basename}.cds_translation.{{cds}}"));
+      let output_translations_path = add_extension(output_translations_path, "fasta");
 
       let output_translations_template = output_translations_path
         .to_str()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## _NEXT_
+
+### CLI
+
+#### Bug fixes
+
+- Fixed a bug introduce in v3.0.0 which caused the default path for translations to be incorrect. This affected only users who used `--output-all` without passing a custom path template via `--output-translations`. The new default path is `nextclade.cds_translation.{cds}.fasta` where `{cds}` gets replaced with the name of the CDS, e.g. `nextclade.cds_translation.S.fasta` for SARS-CoV-2's spike protein.
+
+#### Documentation
+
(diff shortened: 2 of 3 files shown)


packages_rs/nextclade-cli/src/cli/nextclade_cli.rs
```diff
@@ -835,7 +835,6 @@ pub fn nextclade_get_output_filenames(run_args: &mut NextcladeRunArgs) -> Result
     if output_selection.contains(&NextcladeOutputSelection::Translations) {
       let output_translations_path =
         default_output_file_path.with_file_name(format!("{output_basename}.cds_translation.{{cds}}.fasta"));
-      let output_translations_path = add_extension(output_translations_path, "translation.fasta");
```
ivan-aksamentov, line 838: I think it's better to keep `add_extension()` call, just change both the strings.

I no longer remember details anymore, but there were weirdities when filename contains dots: https://github.com/nextstrain/nextclade/issues/907

This is the original (erroneous) commit: https://github.com/nextstrain/nextclade/commit/aea4cde536c5c7624c3d0170fe4b5a479b5bbd2f

ivan-aksamentov:
Regarding breakage, I am not sure. Bug fixes need to be done eventually, so do you consider a fix to also be a breakage or not? I mean we could ship v4 technically and very easily, but incrementing the version will help noone. So I'd just release it as a bug fix.

And, yes, it should be documented. If not, then it's a bug in the docs.

corneliusroemer:
Totally agreed that this is a fix and not a breaking change re v3.0.0, so should be 3.0.1

Sounds good to add to migration guide and 3.0.1 release notes

packages_rs/nextclade-cli/src/cli/nextclade_cli.rs
corneliusroemer: I'll see whether I can figure it out, thanks for the pointer to prevent a regression!

corneliusroemer: This PR currently doesn't have the issue reported in #907 (i.e. it doesn't regress):
```
❯ cargo run --bin nextclade -- run -d sars-cov-2 --output-all out --output-basename basename.THISWILLDISAPPEAR -t out/test.tsv -- data/sars-cov-2/sequences.fasta
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
     Running `target/debug/nextclade run -d sars-cov-2 --output-all out --output-basename basename.THISWILLDISAPPEAR -t out/test.tsv -- data/sars-cov-2/sequences.fasta`

nextclade_master on  fix/output-translation-path [$!?] via  v20.5.1 via 🐍 v3.11.7 via 🦀 v1.70.0 via 🅒 py11 took 5s 
❯ ls out
basename.THISWILLDISAPPEAR.aligned.fasta                basename.THISWILLDISAPPEAR.cds_translation.ORF7b.fasta
basename.THISWILLDISAPPEAR.auspice.json                 basename.THISWILLDISAPPEAR.cds_translation.ORF8.fasta
basename.THISWILLDISAPPEAR.cds_translation.E.fasta      basename.THISWILLDISAPPEAR.cds_translation.ORF9b.fasta
basename.THISWILLDISAPPEAR.cds_translation.M.fasta      basename.THISWILLDISAPPEAR.cds_translation.S.fasta
basename.THISWILLDISAPPEAR.cds_translation.N.fasta      basename.THISWILLDISAPPEAR.csv
basename.THISWILLDISAPPEAR.cds_translation.ORF1a.fasta  basename.THISWILLDISAPPEAR.json
basename.THISWILLDISAPPEAR.cds_translation.ORF1b.fasta  basename.THISWILLDISAPPEAR.ndjson
basename.THISWILLDISAPPEAR.cds_translation.ORF3a.fasta  basename.THISWILLDISAPPEAR.nwk
basename.THISWILLDISAPPEAR.cds_translation.ORF6.fasta   test.tsv
basename.THISWILLDISAPPEAR.cds_translation.ORF7a.fasta
```

corneliusroemer: Done in 56b04356bb351a355a73e47137b7a8b47dc9d859

Didn't observe any change but better be safe :)

Now good?

ivan-aksamentov (approved):

Issue #1388, opened 2024-01-23
PCR primer mutation functionality[v3]


sureman:
This is a request to restore the --input-pcr-primers functionality in CLI v3 as we do use this in our pipelines.

corneliusroemer:
Thanks for reaching out with this feedback! The functionality is still present, what's changed is how to pass the config (in v3 as JSON as part of `--input-pathogen-json` instead of as a csv with its own cli argument). Could you let us know how you're using the CLI flag and the feature? 

We weren't aware of anyone using this feature with custom primers, if you could give a bit more context I'm sure we can figure out a way. It's likely already possible with v3.0.0 what you'd want. 

Good reminder for me complete that part of the v3 documentation.

sureman:
Thank you for responding so quickly.  We were using the --input-pcr-primers parameter to input a VarSkip primer tsv such as:

Country (Institute) | Target | Oligonucleotide | Sequence
-- | -- | -- | --
varskip_neb_vss2b | nCoV-2019_1 | varskip-0317-1_1_LEFT | GGTAACAAACCAACCAACTTTCGA

Given this file, NextClade would return the primer, mutation and genome position within the pcrPrimerChanges field of the report if detected.

corneliusroemer:
Thanks for the quick reply @sureman, that's useful! Good to see the feature used for sequencing primers!

To reassure you, the PCR primer feature hasn't been removed, just the way the config file is passed has. I will document the how-to. For now, if the feature is important to your use case, there's nothing wrong with continuing to use v2 for the moment. There's no degradation of v2 in any way, datasets still work the same etc (though there likely won't be new datasets, but for SARS-CoV-2 this will only happen in 4-6 weeks anyways), so we have a bit of time.

The current v3.0.0 way of doing things would involve (if I understand your usage correctly):
1. Downloading the dataset you would like to use, e.g. `sars-cov-2`, putting it into a folder
2. Patching the `pathogen.json` file by adding your primer details at the appropriate path (e.g. a short Python script that takes a csv or bed file and the downloaded pathogen json  and outputs the patched/customized pathogen json
3. Run nextclade v3 with the downloaded dataset except for the fact that you overwrite the `--input-pathogen-json` with the patched one

Just so I have the full context, could you maybe tell me a bit more about:
- which dataset you usually use
- what other CLI arguments you use

corneliusroemer:
[...]

PR #1398, opened 2024-02-01
feat(cli): bring back PCR primers

ivan-aksamentov:
Resolves: https://github.com/nextstrain/nextclade/issues/1388

Brings back a possibility to use`primers.csv` file and calculate PCR primer mutations in Nextclade CLI v3, just like in Nextclade v2.

What's done:

- [x] add `--input-pcr-primers` CLI argument, accepting a path to input `primers.csv`file in the same format as it was in Nextclade v2 ([example](https://github.com/nextstrain/nextclade_data/blob/e9a61fa38ac956b1ea3087aede1921b7d4ecb8af/data/datasets/sars-cov-2/references/MN908947/versions/2024-01-15T12%3A00%3A00Z/files/primers.csv))
- [x] add parser for`primers.csv`file
- [x] calculate PCR primers mutations based on the `primers.csv`file
- [x] emit PCR primer mutations into output files, just like in Nextclade v2
- [x] ignore `primers` field in `pathogen.json` - the JSON format was such that it is unrealistic to produce it, and it is not used anywhere

What it does not do:
- unlike Nextclade v2, this version does not read `primers.csv` file from a dataset
- Nextclade Web does not use primers at all - an empty array of primers is always passed into the algorithm


corneliusroemer (approved):
Tested with v2 primer csv and it worked. I got expected output in the tsv.

PR #1400, opened 2024-02-01
fix(web): ensure table stripes are always alternating

ivan-aksamentov:
Ensure results table row colors always alternating and look like zebra stripes, regardless of sorting or filtering applied to the rows.


corneliusroemer (approved):
Nice, tested and works as expected. Low risk change, if something happens I will notice on master very quickly

PR #1409, opened 2024-02-11
WIP: reduce minimizer threshold from 0.3 to 0.1

rneher:
for very diverse viruses, our current match fraction threshold of 0.3 for minimizers is too stringent. 0.1 is still very sensitive in the sense that no random hits are produced. But we have many suboptimal hits between related viruses. (like RSV-A matching RSV-B). This is not a problem as long as we consider the best hit. But in the `sort` command we end up not sorting anymore.


rneher:
The sort command currently outputs a sequence into a file corresponding to the dataset prefix if a dataset with this prefix is a hit. If we changed this to the more stringent condition that the dataset needs to be the *best* hit, the sort would work more as a hierarchical sort.

rneher:
If I understand this correctly, one could only process the first (best) dataset here:
https://github.com/nextstrain/nextclade/blob/master/packages_rs/nextclade-cli/src/cli/nextclade_seq_sort.rs#L200-L207

Issue #1422, opened 2024-02-20
Empty input file causes uncaught error in v3 (it didn't in v2)


corneliusroemer:
Migrating ncov-ingest to v3, I noticed that v3 errors on empty input fasta file:

```shell
$ nextclade3 run test.fasta -d sars-cov-2 -t test.tsv
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: 
   0: Expected character '>' at record start.

Location:
   packages/nextclade/src/io/fasta.rs:106

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
Location: packages/nextclade-cli/src/cli/nextclade_loop.rs:80

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
zsh: abort      nextclade3 run test.fasta -d sars-cov-2 -t test.tsv
```

In v2, the behaviour was to exit with exit code 0 and without anything written to stderr:

```shell
$ nextclade2 run test.fasta -d sars-cov-2 -t test.tsv                                                                    


```

Tested on v3.2.1

PR #1424, opened 2024-02-23
fix: handle leading newlines in fasta parser correctly 

ivan-aksamentov:
Resolves https://github.com/nextstrain/nextclade/issues/1422

`FastaReader` throws an error if there's a leading newline in the inputs.

Because of the changes in `Concat` iterator in Nextclade v3, `FastaReader` now receives additional newlines, which `Concat` inserts between input readers (e.g. files or string streams). This was a necessary fix to ensure that there's a separation between two concatenated readers when the previous reader does not end with a newline. However, now if the first reader (or the only reader)  is empty or reads only newlines, then after `Concat` `FastaReader` receives only a `\n` characters, which it could not interpret and fails with an error. This is a change of behavior compared to Nextclade v2.

In the new version I make sure to skip the leading newlines. The danger of this version is the `loop` block. The termination of this loop entirely depends on the inputs, so it is important to find all corner cases to avoid infinite loop and to avoid terminating it too early.

### DIFF
--- a/packages/nextclade/src/io/concat.rs
+++ b/packages/nextclade/src/io/concat.rs
@@ -33,6 +33,15 @@ where
 
   /// Concatenate readers into a single reader, alternating them with the provided delimiter,
   /// i.e. inserting this sequence of bytes between adjacent readers.
+  ///
+  /// The idea is that if readers (e.g. files) don't contain trailing newline, it will make line parsers behave
+  /// incorrectly on reader boundaries. So one can add newline delimiter between readers, to compensate for that.
+  ///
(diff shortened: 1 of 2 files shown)


ivan-aksamentov:
Hm... I also have an alternative idea. Instead of concatenating the input readers, we could simply iterate over them and create a fresh `FastaReader` instance for each one. This way there is no need to insert any delimiters between the readers, and so no need to handle extra newlines. This might work even better.

Then we may or may not preserve the support of FASTA files with trailing newlines introduced in this PR. But it will no longer be a requirement for having multiple readers.

PR #1430, opened 2024-03-11
chore(ci): update linux-aarch64 sha in bioconda publish script

corneliusroemer:
Bioconda now publishes linux-aarch64 packages for nextclade as well, so those shas also should be updated.

https://github.com/bioconda/bioconda-recipes/pull/46185/

### DIFF
--- a/scripts/publish_bioconda
+++ b/scripts/publish_bioconda
@@ -29,6 +29,7 @@ export artifacts_dir="${THIS_DIR}/../artifacts"
 
 artifacts=(
   'nextclade-aarch64-apple-darwin'
+  'nextclade-aarch64-unknown-linux-gnu'
   'nextclade-x86_64-apple-darwin'
   'nextclade-x86_64-unknown-linux-gnu'
 )


ivan-aksamentov:
Is there no need to also update https://github.com/nextstrain/nextclade/blob/514c5c50ef8f4c58f542cb2abbfbf967370b9268/scripts/update-bioconda.py? I don't quite remember how this works. Upon quick look, I think the change of selector comments like `[arm64]` in the `meta.yaml` in the bioconda repo breaks this script.

corneliusroemer:
Yes that script would also no longer work, good spot! I don't know whether we still use that at all though.

I think that arm64 === osx and arm64, i.e. the osx is unnecessary but I thought I make the os explicit as well as it's not super obvious that aarch64 is Linux and arm64 is Mac.

ivan-aksamentov:
>  I don't know whether we still use that at all though.

We definitely use this script, it does most of the work:
https://github.com/nextstrain/nextclade/blob/514c5c50ef8f4c58f542cb2abbfbf967370b9268/scripts/publish_bioconda#L82-L83


> aarch64 is Linux and arm64 is Mac

That's weird.

"AArch64 and "ARM64" are synonyms. The official name of the architecture is "AArch64" and "ARM64" is a colloquial name (probably to make it sound similar to "AMD64"). There is no "ARM for Linux" and "ARM for Mac", these are all flavors of the same architecture licensed from ARM, and you can install any software on any hardware if you try hard enough (e.g. https://asahilinux.org), barring the artificial limitations by Apple.

Are these comments documented somewhere? Where do they come from? Wouldn't it be wiser to use generally accepted target triplet nomenclature instead of inventing random stuff?

corneliusroemer:
Ah I missed that, thanks!

Yes they are documented here: https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#preprocessing-selectors
selector | description
-- | --
x86 | True if the system architecture is x86, both 32-bit and 64-bit, for Intel or AMD chips.
x86_64 | True if the system architecture is x86_64, which is 64-bit, for Intel or AMD chips.
linux | True if the platform is Linux.
linux32 | True if the platform is Linux and the Python architecture is 32-bit and uses x86.
linux64 | True if the platform is Linux and the Python architecture is 64-bit and uses x86.
armv6l | True if the platform is Linux and the Python architecture is armv6l.
armv7l | True if the platform is Linux and the Python architecture is armv7l.
aarch64 | True if the platform is Linux and the Python architecture is aarch64.
ppc64le | True if the platform is Linux and the Python architecture is ppc64le.
s390x | True if the platform is Linux and the Python architecture is s390x.
osx | True if the platform is macOS.
arm64 | True if the platform is either macOS or Windows and the Python architecture is arm64.
unix | True if the platform is either macOS or Linux or emscripten.
win | True if the platform is Windows.
win32 | True if the platform is Windows and the Python architecture is 32-bit.
win64 | True if the platform is Windows and the Python architecture is 64-bit.
py | The Python version as an int, such as 27 or 36. See the CONDA_PY environment variable.
py3k | True if the Python major version is 3.
py2k | True if the Python major version is 2.
py27 | True if the Python version is 2.7. Use of this selector is discouraged in favor of comparison operators (e.g. py==27).
py34 | True if the Python version is 3.4. Use of this selector is discouraged in favor of comparison operators (e.g. py==34).
py35 | True if the Python version is 3.5. Use of this selector is discouraged in favor of comparison operators (e.g. py==35).
py36 | True if the Python version is 3.6. Use of this selector is discouraged in favor of comparison operators (e.g. py==36).
np | The NumPy version as an integer such as 111. See the CONDA_NPY environment variable.
build_platform | The native subdir of the conda executable
```

They screwed up the separation of os and arch when they made `aarch64` being necessarily linux, and `arm64` being not linux (either windows or osx).

We can be explicit by using redundancy:
- `linux and x86`
- `linux and aarch64`
- `osx and x86`
- `osx and arm64`

We _have_ to use different terms for arm64/aarch64 depending on platform though - it's annoying :/

corneliusroemer:
Oh no, I have to fix this PR here otherwise we will not update it for aarch64 🙃

corneliusroemer:
Closed in favor of #1434

Issue #1437, opened 2024-04-17
Scrollbar shown for dataset names in dataset picker


corneliusroemer:
There's a tiny visual bug in v3 dataset picker: a scrollbar is shown for all dataset names as it seems that the height of the field is too small for the font size:

![2024-04-17 18 17 23](https://github.com/nextstrain/nextclade/assets/25161793/53dc2482-e93c-4516-8651-01f0169535bc)

ivan-aksamentov:
Wow! That's bad. Any way to repro this for me?

Can you check it for me in dev tools? Right click on the text -> Inspect -> find `h4` element nearby in dev tools, click on it -> In sidebar "Computed" -> "Rendered Fonts"

What does it show?

corneliusroemer:
Like this?

macOS 14.4.1, Brave v 1.64.122

<img width="1105" alt="image" src="https://github.com/nextstrain/nextclade/assets/25161793/e203c5bc-5d73-4cb9-8b34-fa82300938a7">

ivan-aksamentov:
When did that happen? Was it always the case? I haven't touched web in weeks.

ivan-aksamentov:
Are you using zoom? How does it look without zoom?

corneliusroemer:
> When did that happen? Was it always the case? I haven't touched web in weeks.

Yes it's been the case for me since the redesign for v3 - I think I might have mentioned it in some slack thread but because it's only minor visual I didn't make an issue and I just got used to it :)

> Are you using zoom? How does it look without zoom?

The issue is present across _all_ zoom levels

![2024-04-17 18 50 10](https://github.com/nextstrain/nextclade/assets/25161793/58a0bf4f-00db-4224-bb27-70a75cd1c816)

corneliusroemer:
If you tell me what to tweak I can give it a try!

corneliusroemer:
It appears to be fixed!

<img width="1485" alt="image" src="https://github.com/nextstrain/nextclade/assets/25161793/5c220987-d8c6-4327-a5c7-912bc27b6cc0">

ivan-aksamentov:
This has been fixed in https://github.com/nextstrain/nextclade/pull/1438 and released in [3.4.0](https://github.com/nextstrain/nextclade/releases/tag/3.4.0)

PR #1438, opened 2024-04-17
fix(web): attempt to remove vertical scrollbars in dataset  names

ivan-aksamentov:
https://github.com/nextstrain/nextclade/issues/1437


corneliusroemer:
It works, but cuts off the things below the line, e.g. the low part of the `p` but I think that's fine, and is what happened anyways - now we just get rid of scroll bar.

image

PR #1440, opened 2024-04-19
feat(web): select top dataset when suggest was triggered manually

ivan-aksamentov:
This changes behavior of dataset suggestions: now when use clicks "Suggest" button on main page, if suggestion run completes successfully, then top matching dataset is selected as current dataset automatically. This does not happen after automatic suggestion runs.


ivan-aksamentov:
Currently broken: automatic selection only happens after second click when autosuggest toggle is disabled.

PR #1446, opened 2024-05-08
fix: ensure loss of aa motif detected when it's the only motif

ivan-aksamentov:
Currently Nextclade fails to detect a lost motif if the motif is the only one in its category. This is due to incorrect iteration of the category keys (names) of motifs.

Here I implemented an iterator to correctly visit both the reference and query motif maps: it will visit the category even if it is entirely missing in one or the other map. Previously, the category missing from one map would not be considered, which caused silent failure to detect the disappearance of a motif.

This was initially reported for the flu h5 datasets - the loss of `polybasic_cleavage_site` motif would not be detected.

I tested by comparing the outputs of smoke tests from master branch and from this branch. The `polybasic_cleavage_site` motif loss is now detected in h5 datasets and there are no other changes.


rneher:
thanks. looks good to me.

Issue #1451, opened 2024-05-10
Developer guide uses deprecated CLI option


genehack:
In a couple places, the developer guide has example invocations that use `--output-dir`. Running them results in the following error message: 

```
The argument `--output-dir` is removed in favor of `--output-all`.
```

PR #1452, opened 2024-05-10
Change output-dir to output-all

genehack:
Updates developer guide to replace deprecated option with supported one. 

#1451


rneher:
I don't think the first change is correct. The `dataset get` command expects `--output-dir`. the second one looks right.

genehack:
> I don't think the first change is correct. The `dataset get` command expects `--output-dir`. the second one looks right.

Ope! You are correct. I pushed an update to roll the first one back.

ivan-aksamentov (approved):

PR #1454, opened 2024-05-14
feat: mutations relative to arbitrary node

ivan-aksamentov:
This extends concept of private mutations (relative to the parent node on the ref tree) to mutations relative to an arbitrary node of interest.

The ref nodes of interest are described by the user in the `.meta .extensions .nextclade .reference_nodes` of the input Auspice JSON. The description can also contain constrains: we can match node to only query samples belonging to a certain clade or lineage.

Private mutations functionality is unchanged. New functionality, inputs and outputs are added. Though the implementation algo is largely reused.

### Work items

 - [x] read input config from Auspice JSON
 - [x] calculate relative nuc mutations
 - [x] calculate relative aa mutations
 - [x] filter by clade and clade-like attributes
 - [ ] filter by gene?
 - [x] output to Nextclade JSON
 - [x] output to Nextclade NDJSON
 - [ ] output to TSV and CSV
 - [ ] display in web app 

### Inputs

Example configuration object. Put it into `.meta.extensions.nextclade.reference_nodes`

<details>
  <summary>Click to expand</summary>

```json
{
  "extensions": {
    "nextclade": {
      "reference_nodes": [
        {
          "name": "NODE_0000659",
          "displayName": "BA.2.86 (23I)",
          "description": "Ancestral BA.2.86 sequence"
        },
        {
          "name": "XBB.1.5",
          "displayName": "XBB.1.5 (23A)",
          "description": "Ancestral XBB.1.5 sequence. Vaccine strain 2023/2024",
          "include": {
            "clade": ["23A"]
          }
        },
        {
          "name": "NODE_0000862",
          "displayName": "BA.5 (22B)",
          "description": "Ancestral BA.5 sequence. Vaccine strain 2022/2023",
          "include": {
            "clade": ["22B"]
          }
        }
      ]
    }
  }
}
```
</details>

The `name` field should match the `name` field of one of the nodes on the tree. The `displayName` and `description` are optional arbitrary strings used for display purposes.

The `include` field should be an object, which contains:
 - keys: `name`s from the `.meta.extensions.nextclade.clade_node_attrs` (for clade-line attributes) or string `"clade"` (for the built-in clades).
 - values: a list of values of the clade-like attribute or a list of built-in clades. Only query sequences which match these attributes are considered for calculation of mutations relative to that node.

If the `include` field is not present, then no constraints applied (all query sequences are considered).

#### Output JSON and NDJSON

Example fragment of output json entry (entry in the `.results[]` array) (mutation lists are truncated for demonstration purposes)

<details>
  <summary>Click to expand</summary>

```json
{
  "relativeNucMutations": [
    {
      "refNode": {
        "name": "NODE_0000659",
        "displayName": "BA.2.86 (23I)",
        "description": "Ancestral BA.2.86 sequence"
      },
      "muts": {
        "privateSubstitutions": [
          {"pos": 404, "refNuc": "A", "qryNuc": "G"},
          {"pos": 896, "refNuc": "A", "qryNuc": "C"}
        ],
        "privateDeletions": [],
        "reversionSubstitutions": [
          {"pos": 896, "refNuc": "A", "qryNuc": "C"},
          {"pos": 3430, "refNuc": "T", "qryNuc": "G"}
        ],
        "labeledSubstitutions": [
          {
            "substitution": {"pos": 404, "refNuc": "A", "qryNuc": "G"},
            "labels": ["23A", "23D", "23F", "23B", "22F", "23E", "23H", "23G"]
          },
          {
            "substitution": {"pos": 2333, "refNuc": "C", "qryNuc": "T"},
            "labels": ["23F", "23H"]
          }
        ],
        "unlabeledSubstitutions": [
          {"pos": 4089, "refNuc": "C", "qryNuc": "T"},
          {"pos": 11344, "refNuc": "C", "qryNuc": "T"}
        ],
        "totalPrivateSubstitutions": 75,
        "totalPrivateDeletions": 0,
        "totalReversionSubstitutions": 37,
        "totalLabeledSubstitutions": 34,
        "totalUnlabeledSubstitutions": 4
      }
    }
  ],
  "relativeAaMutations": [
    {
      "refNode": {
        "name": "NODE_0000659",
        "displayName": "BA.2.86 (23I)",
        "description": "Ancestral BA.2.86 sequence"
      },
      "muts": {
        "E": {
          "privateSubstitutions": [
            {"cdsName": "E", "pos": 10, "refAa": "T", "qryAa": "A"}
          ],
          "privateDeletions": [],
          "reversionSubstitutions": [],
          "totalPrivateSubstitutions": 1,
          "totalPrivateDeletions": 0,
          "totalReversionSubstitutions": 0
        },
        "S": {
          "privateSubstitutions": [
            {"cdsName": "S", "pos": 20, "refAa": "T", "qryAa": "R"},
            {"cdsName": "S", "pos": 26, "refAa": "-", "qryAa": "S"}
          ],
          "privateDeletions": [
            {"cdsName": "S", "pos": 23, "refAa": "S"},
            {"cdsName": "S", "pos": 143, "refAa": "Y"}
          ],
          "reversionSubstitutions": [
            {"cdsName": "S", "pos": 20, "refAa": "T", "qryAa": "R"},
            {"cdsName": "S", "pos": 49, "refAa": "L", "qryAa": "S"}
          ],
          "totalPrivateSubstitutions": 39,
          "totalPrivateDeletions": 2,
          "totalReversionSubstitutions": 25
        }
      }
    }
  ]
}

```
</details>

#### Output TSV and CSV

TODO

### Visualization in Nextclade Web

TODO

### DIFF
--- a/packages/nextclade/src/analyze/find_relative_mutations.rs
+++ b/packages/nextclade/src/analyze/find_relative_mutations.rs
@@ -1,10 +1,15 @@
 use crate::alphabet::nuc::Nuc;
+use crate::analyze::aa_del::AaDel;
+use crate::analyze::aa_sub::AaSub;
+use crate::analyze::find_private_aa_mutations::{find_private_aa_mutations, PrivateAaMutations};
 use crate::analyze::find_private_nuc_mutations::{find_private_nuc_mutations, PrivateNucMutations};
-use crate::analyze::letter_ranges::NucRange;
+use crate::analyze::letter_ranges::{CdsAaRange, NucRange};
 use crate::analyze::nuc_del::NucDelRange;
 use crate::analyze::nuc_sub::NucSub;
 use crate::analyze::virus_properties::VirusProperties;
-use crate::coord::range::NucRefGlobalRange;
+use crate::coord::range::{AaRefRange, NucRefGlobalRange};
+use crate::gene::gene_map::GeneMap;
+use crate::translate::translate_genes::Translation;
 use crate::tree::tree::{AuspiceGraph, AuspiceRefNode};
 use eyre::{eyre, Report, WrapErr};
 use itertools::Itertools;
@@ -18,7 +23,7 @@ pub struct RelativeNucMutations {
   muts: PrivateNucMutations,
 }
 
-pub fn find_relative_mutations(
+pub fn find_relative_nuc_mutations(
   graph: &AuspiceGraph,
   clade: &str,
   clade_node_attrs: &BTreeMap<String, String>,
@@ -57,7 +62,54 @@ pub fn find_relative_mutations(
       })
     })
     .collect::<Result<Vec<RelativeNucMutations>, Report>>()
-    .wrap_err("When calling mutations relative to a reference node")
+    .wrap_err("When calling nucleotide mutations relative to a reference node")
+}
+
+#[derive(Clone, Serialize, Deserialize, schemars::JsonSchema, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct RelativeAaMutations {
+  ref_node: AuspiceRefNode,
+  muts: BTreeMap<String, PrivateAaMutations>,
+}
+
+pub fn find_relative_aa_mutations(
+  graph: &AuspiceGraph,
+  clade: &str,
+  clade_node_attrs: &BTreeMap<String, String>,
+  aa_substitutions: &[AaSub],
+  aa_deletions: &[AaDel],
+  aa_unknowns: &[CdsAaRange],
+  aa_unsequenced_ranges: &BTreeMap<String, Vec<AaRefRange>>,
+  ref_peptides: &Translation,
+  gene_map: &GeneMap,
+) -> Result<Vec<RelativeAaMutations>, Report> {
+  let ref_nodes = filter_ref_nodes(graph, clade, clade_node_attrs);
+
+  ref_nodes
+    .iter()
+    .map(|&ref_node| -> Result<_, Report> {
+      let node = graph
+        .iter_nodes()
+        .find(|node| node.payload().name == ref_node.name)
+        .ok_or_else(|| eyre!("Unable to find reference node on the tree: '{}'", &ref_node.name))?;
+
+      let muts = find_private_aa_mutations(
+        node.payload(),
+        aa_substitutions,
+        aa_deletions,
+        aa_unknowns,
+        aa_unsequenced_ranges,
+        ref_peptides,
+        gene_map,
+      );
+
+      Ok(RelativeAaMutations {
+        ref_node: ref_node.to_owned(),
+        muts,
+      })
+    })
+    .collect::<Result<Vec<RelativeAaMutations>, Report>>()
+    .wrap_err("When calling amino acid mutations relative to a reference node")
 }
 
 /// Take only ref nodes which are relevant for this sample, according to the node's include list
--- a/packages/nextclade/src/analyze/find_relative_mutations.rs
+++ b/packages/nextclade/src/analyze/find_relative_mutations.rs
@@ -1,0 +1,85 @@
+use crate::alphabet::nuc::Nuc;
+use crate::analyze::find_private_nuc_mutations::{find_private_nuc_mutations, PrivateNucMutations};
+use crate::analyze::letter_ranges::NucRange;
+use crate::analyze::nuc_del::NucDelRange;
+use crate::analyze::nuc_sub::NucSub;
+use crate::analyze::virus_properties::VirusProperties;
+use crate::coord::range::NucRefGlobalRange;
+use crate::tree::tree::{AuspiceGraph, AuspiceRefNode};
+use eyre::{eyre, Report, WrapErr};
+use itertools::Itertools;
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
+
+#[derive(Clone, Serialize, Deserialize, schemars::JsonSchema, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct RelativeNucMutations {
+  ref_node: AuspiceRefNode,
+  muts: PrivateNucMutations,
+}
+
+pub fn find_relative_mutations(
+  graph: &AuspiceGraph,
+  clade: &str,
+  clade_node_attrs: &BTreeMap<String, String>,
+  substitutions: &[NucSub],
+  deletions: &[NucDelRange],
+  missing: &[NucRange],
+  alignment_range: &NucRefGlobalRange,
+  ref_seq: &[Nuc],
+  non_acgtns: &[NucRange],
+  virus_properties: &VirusProperties,
+) -> Result<Vec<RelativeNucMutations>, Report> {
+  let ref_nodes = filter_ref_nodes(graph, clade, clade_node_attrs);
+
+  ref_nodes
+    .iter()
+    .map(|&ref_node| -> Result<_, Report> {
+      let node = graph
+        .iter_nodes()
+        .find(|node| node.payload().name == ref_node.name)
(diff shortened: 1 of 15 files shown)


packages/nextclade/src/analyze/find_relative_mutations.rs
```diff
@@ -0,0 +1,137 @@
...
+        ref_seq,
+        non_acgtns,
+        virus_properties,
+      );
+
+      Ok(RelativeNucMutations {
+        ref_node: ref_node.to_owned(),
+        muts,
+      })
+    })
```
ivan-aksamentov, line 63: There is very little new logic, mostly bookkeeping. The `find_private_*_mutations()` for nucs and aa are reused as is. Only difference is that it runs multiple times for all requested nodes.

packages/nextclade/src/analyze/find_relative_mutations.rs
```diff
@@ -0,0 +1,137 @@
...
+    .reference_nodes
+    .iter()
+    .filter(|node| {
+      // For each attribute key in includes, check that the attribute value of this sample match
+      // at least one item in the include list
+      node.include.iter().all(|(key, includes)| {
+        let curr_value = if key == "clade" { clade } else { &clade_node_attrs[key] };
+        includes.iter().any(|include_value| include_value == curr_value) // TODO: consider regex match rather than equality
+      })
+    })
```
ivan-aksamentov, line 135: This is the logic for constraining the mutations calculation by clades and clade-like attributes. If `include` field is present, then we lookup the constrained attribute of the query sample and only consider this node if the query attribute's value is matching any of the values in the include list.

For example, if config has node of interest which is only relevant for clades 23A and 23B:

```json
{
  include: { clade: ["23A", "23B"] }
}
```

then mutations relative to this node will be calculated only for query samples of clade 23A and 23B.


Same for pango lineages:

```json
{
  include: { Nextclade_pango: ["A.1.2.3", "A.1.2.3.4"] }
}
```

It is up for discussion how multiple keys in the `include` object should be handled - using boolean OR and AND.

rneher:
Is the choice of comparison reference persisted in local storage? Because I switch from H3 to H1 and it had an H3 reference selected which doesn't exist (and all sequences were greyed out). 

image

We might want to default always to the original alignment reference.

rneher:
In the `relativeMutations['strain'].deletions` column in the tabular output, there are some instances of `empty range`. This might result from deletions being present in the comparison node and the query sequence. It would be most consistent to not output "emtpy range", I think .

huddlej (commented):
This is really excellent, @ivan-aksamentov and @rneher! I had just been thinking about how to best generate [the “derived haplotype” annotations](https://nextstrain.org/seasonal-flu/h3n2/ha/2y?branchLabel=Subclade&c=haplotype&d=tree&label=Subclade:J.2&p=full) in our seasonal flu builds for all sequences from the Nextclade TSV alone and this new feature seems to offer exactly that functionality.

Ideally there could be an option in the dropdown to compare to “nearest clade” instead of “parent” (traversing from each tip to the nearest ancestral node with a clade label). This functionality would eliminate the need for each Nextclade dataset to define reference_nodes entries for each clade, since that mapping of clade name to node in tree and sequence already exists in the dataset. In the TSV/CSV output, these clade-specific substitutions could still appears in separate columns named after the clade labels from the tree (e.g., relativeMutations['J.2'].aaSubstitutions). One potential issue with this feature would be the presence of multiple options for “clade” columns (e.g., “Clade” and “Subclade” for seasonal flu) that the dataset developer would need to tune. Maybe this is why you went with the manually defined clades in the extensions approach you did?

In practice, it seems that we can add `extensions` to the Auspice JSON either by defining the extensions in the Auspice config JSON we pass to the `augur export v2` command or by modifying the Auspice JSON with a custom script after it has been created by the export command. The example inputs in this PR reference internal node names of the tree (`NODE_0000659`) which are values that can change with each new run of the Nextclade workflow. The example input also shows one node name of `XBB.1.5` which seems to match a clade label defined in the branch attributes for a given node. Is it possible for us to define our inputs with these more generic clade names instead of the specific variable tree node names? For example, could I define an H3N2 HA config like the following to get mutations relative to subclades J.1 and J.2?

```json
{
  "extensions": {
    "nextclade": {
      "reference_nodes": [
        {
          "name": "J.1",
          "include": {
            "subclade": ["J.1"]
          }
        },
        {
          "name": "J.2",
          "include": {
            "subclade": ["J.2"]
          }
        }
      ]
    }
  }
}
```

Or would I need to write a script to create the `extensions` entries above from the Auspice JSON I get from `augur export`, so I can map specific node names to clade names?

ivan-aksamentov:
@huddlej 

> Is it possible for us to define our inputs with these more generic clade names instead of the specific variable tree node names? For example, could I define an H3N2 HA config like the following to get mutations relative to subclades J.1 and J.2?

Makes sense! Should not be too hard technically. But I don't understand consequences of this more broadly, so I'll rely on yours and Richards guidance here. I guess you two will be the primary target users of this json format, so let's make it nice and convenient.

The current name matching logic is [here](https://github.com/nextstrain/nextclade/blob/6c6cf1ce0a1f054f30a541f96cc7024053f48f80/packages/nextclade/src/analyze/find_relative_mutations.rs#L97). It's a matter of changing this conditional. Right now it matches a string property at `node.name` with the `ref_node.name` which comes from the description json snippet. What exact node property path you propose instead?

We can also have a more complex description object format and match either `node.name` or something else. The struct which mirrors the description json object is [here](https://github.com/nextstrain/nextclade/blob/6c6cf1ce0a1f054f30a541f96cc7024053f48f80/packages/nextclade/src/analyze/find_relative_mutations.rs#L97).

---

Additionally to this, we weren't sure about the clade filtering mechanism either (the `include` field). Do you want to always include only the same clade as the node of interest? Or perhaps this clade plus all newer clades? Maybe it makes sense to actually have an `exclude` mechanism instead (or additional to). The include mechanism is currently implemented [like this](https://github.com/nextstrain/nextclade/blob/6c6cf1ce0a1f054f30a541f96cc7024053f48f80/packages/nextclade/src/analyze/find_relative_mutations.rs#L123-L146). Another improvement could be to use regex match rather than string equality.

Would be nice to discuss the file format thoroughly now and only ship something final, because it will be difficult to change this without breaking lots of stuff later.

ivan-aksamentov:
> Maybe this is why you went with the manually defined clades in the extensions approach you did?

It could be! (Richard is the brainpower behind this feature :)) But more likely is that we simply did not think about it much. This is the first working prototype, and the simplest form we could think of. But we might very well change how it works if you'd like to propose changes!

ivan-aksamentov:
@rneher 

> Is the choice of comparison reference persisted in local storage? Because I switch from H3 to H1 and it had an H3 reference selected which doesn't exist (and all sequences were greyed out).

 - [x] Bug with persistent node selection is fixed in [`8cedad4` (#1454)](https://github.com/nextstrain/nextclade/pull/1454/commits/8cedad4e26ae5d10285235c8f6c1d8718feb7392).

- [x] "Reference" "node" is now default (instead of "parent") [`e8c2160` (#1454)](https://github.com/nextstrain/nextclade/pull/1454/commits/e8c2160babf62ee1c672ba4ad08b084e482f0fb7)

> In the relativeMutations['strain'].deletions column in the tabular output, there are some instances of empty range. This might result from deletions being present in the comparison node and the query sequence. It would be most consistent to not output "emtpy range", I think 

 - [x] bug with incorrect grouping of individual deletions into ranges is fixed in [`149f294` (#1454)](https://github.com/nextstrain/nextclade/pull/1454/commits/149f294768401bab7c69bf3452029c6247b88a25) - empty ranges should never be produced

ivan-aksamentov:
I thought that in the description json we could also optionally define the default node of interest and it will be preselected in the dropdown in web. Similarly to how it's done for the cds dropdown.

PR #1460, opened 2024-05-30
feat: fetch sidecar Auspice JSON if `.root_sequence` is not on the tree

ivan-aksamentov:
Followup of https://github.com/nextstrain/nextclade/pull/1455

If `.root_sequence` is not available on Auspice JSON, let's attempt to fetch ref sequence from sidecar Auspice JSON. For that let's GET from the same URL, but with `Accept: application/vnd.nextstrain.dataset.root-sequence+json` header.

### DIFF
--- a/packages/nextclade-web/src/io/fetchSingleDatasetAuspice.ts
+++ b/packages/nextclade-web/src/io/fetchSingleDatasetAuspice.ts
@@ -7,7 +7,7 @@ export async function fetchSingleDatasetAuspice(datasetJsonUrl_: string) {
   const datasetJsonUrl = removeTrailingSlash(datasetJsonUrl_)
 
   const auspiceJson = await axiosFetch<AuspiceTree>(datasetJsonUrl, {
-    headers: { Accept: 'application/json, text/plain, */*' },
+    headers: { Accept: 'application/vnd.nextstrain.dataset.main+json;q=1, application/json;q=0.9, text/plain;q=0.8, */*;q=0.1' },
   })
 
   if (isEmpty(auspiceJson.root_sequence)) {
--- a/packages/nextclade-web/src/io/fetchSingleDatasetAuspice.ts
+++ b/packages/nextclade-web/src/io/fetchSingleDatasetAuspice.ts
@@ -7,7 +7,9 @@ export async function fetchSingleDatasetAuspice(datasetJsonUrl_: string) {
   const datasetJsonUrl = removeTrailingSlash(datasetJsonUrl_)
 
   const auspiceJson = await axiosFetch<AuspiceTree>(datasetJsonUrl, {
-    headers: { Accept: 'application/vnd.nextstrain.dataset.main+json;q=1, application/json;q=0.9, text/plain;q=0.8, */*;q=0.1' },
+    headers: {
+      Accept: 'application/vnd.nextstrain.dataset.main+json;q=1, application/json;q=0.9, text/plain;q=0.8, */*;q=0.1',
+    },
   })
 
   if (isEmpty(auspiceJson.root_sequence)) {
--- a/packages/nextclade-web/src/io/fetchSingleDatasetAuspice.ts
+++ b/packages/nextclade-web/src/io/fetchSingleDatasetAuspice.ts
@@ -1,6 +1,7 @@
+import { isEmpty } from 'lodash'
 import { attrStrMaybe, AuspiceTree, Dataset } from 'src/types'
 import { removeTrailingSlash } from 'src/io/url'
-import { axiosFetch } from 'src/io/axiosFetch'
+import { axiosFetch, axiosFetchOrUndefined } from 'src/io/axiosFetch'
 
 export async function fetchSingleDatasetAuspice(datasetJsonUrl_: string) {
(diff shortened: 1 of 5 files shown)


packages/nextclade-web/src/io/fetchSingleDatasetAuspice.ts
```diff
@@ -1,13 +1,24 @@
...
   })
+
+  if (isEmpty(auspiceJson.root_sequence)) {
+    const sidecar = await axiosFetchOrUndefined<Record<string, string>>(datasetJsonUrl, {
+      headers: { Accept: 'application/vnd.nextstrain.dataset.root-sequence+json' },
+    })
+    if (!isEmpty(sidecar)) {
+      auspiceJson.root_sequence = sidecar
+    }
+  }
```
ivan-aksamentov, line 20: Is this correct to just take the sidecar contents and merge it into `.root_sequence`, like if it always was there?

ivan-aksamentov:
Attempt to run with a very large tree has hit the local storage limit.

https://nextclade-git-feat-fetch-auspice-sidecar-json-nextstrain.vercel.app//?dataset-json-url=https://nextstrain.org/ncov/gisaid/global/all-time

Well, I should not store the tree in the local storage in the first place. Need to rethink my decisions about how to persist the datasets from remote sources (if it all?)

packages/nextclade-web/src/io/fetchSingleDatasetAuspice.ts
jameshadfield: Yup! ([Here's the relevant augur code](https://github.com/nextstrain/augur/blob/f0900bce1a51c4c82b81486b2bcc709db6c08239/augur/export_v2.py#L1239-L1248))

ivan-aksamentov:
It's unclear how to implement a symmetric functionality in CLI. I guess we can tell users to paste their sidecar into the main JSON? Does it make sense to introduce a new CLI arg for a new file type?

packages/nextclade-web/src/io/fetchSingleDatasetAuspice.ts
tsibley: And the [v2 dataset schema](https://nextstrain.org/schemas/dataset/v2) [references the root-sequence sidecar schema](https://github.com/nextstrain/augur/blob/f0900bce1a51c4c82b81486b2bcc709db6c08239/augur/data/schema-export-v2.json#L124-L126) for the `.root_sequence` field.

tsibley:
@ivan-aksamentov The root sequence sidecar for `x_y_z.json` is canonically at `x_y_z_root-sequence.json` when dealing with a file system, so for the CLI it'd seem reasonable to look for that file automatically.

Maybe the CLI could also support fetching a dataset specified as a URL directly, and then make the same content-nego request behind the scenes in that case?

packages/nextclade-web/src/io/fetchSingleDatasetAuspice.ts
```diff
@@ -1,13 +1,24 @@
...
 import { attrStrMaybe, AuspiceTree, Dataset } from 'src/types'
 import { removeTrailingSlash } from 'src/io/url'
-import { axiosFetch } from 'src/io/axiosFetch'
+import { axiosFetch, axiosFetchOrUndefined } from 'src/io/axiosFetch'
 
 export async function fetchSingleDatasetAuspice(datasetJsonUrl_: string) {
   const datasetJsonUrl = removeTrailingSlash(datasetJsonUrl_)
 
   const auspiceJson = await axiosFetch<AuspiceTree>(datasetJsonUrl, {
     headers: { Accept: 'application/json, text/plain, */*' },
```
tsibley, line 10: Since using our media type in the sidecar fetch below, might as well use our main dataset JSON media type here too, with quality factors to rank preferences, e.g.

```suggestion
    headers: { Accept: 'application/vnd.nextstrain.dataset.main+json;q=1, application/json;q=0.9, text/plain;q=0.8, */*;q=0.1' },
```

packages/nextclade-web/src/io/fetchSingleDatasetAuspice.ts
```diff
@@ -1,13 +1,24 @@
...
   })
+
+  if (isEmpty(auspiceJson.root_sequence)) {
+    const sidecar = await axiosFetchOrUndefined<Record<string, string>>(datasetJsonUrl, {
+      headers: { Accept: 'application/vnd.nextstrain.dataset.root-sequence+json' },
```
tsibley, line 15: You might assert here that the response Content-Type is `application/vnd.nextstrain.dataset.root-sequence+json`, so that random JSON isn't shoved into the `root_sequence` field?

For example, consider what happens if the input URL is https://data.nextstrain.org/ncov_open_global_all-time.json...

image

tsibley: Alternatively, you could validate `sidecar` here against https://nextstrain.org/schemas/dataset/root-sequence.

tsibley: LGTM!

image

PR #1461, opened 2024-05-30
refactor: remove recursion when writing out Auspice JSON

ivan-aksamentov:
Recursion has been overflowing call stack in the web app when very large trees are used.

Let's remove recursion when converting internal graph into Auspice JSON when writing out Auspice JSON tree. 

In this version I convert all nodes in advance, and store them in a map (which may or may not be needed) and then explore the tree using stack-based post-order traversal.

This should provide byte-equal JSONs when compared to the previous, recursive implementation.

### DIFF
--- a/packages/nextclade/src/graph/graph.rs
+++ b/packages/nextclade/src/graph/graph.rs
@@ -12,7 +12,7 @@ use itertools::Itertools;
 use num_traits::Float;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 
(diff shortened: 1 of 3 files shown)


ivan-aksamentov:
This was first discovered when using RSV-A Auspice dataset from 2024-05-27 and example sequences from the official RSV-A Nextclade dataset:

Before:
https://nextclade-git-feat-ref-and-ann-from-tree-json-nextstrain.vercel.app/?dataset-json-url=https://nextstrain.org/rsv/a/F@2024-05-27

After:
https://nextclade-git-refactor-auspice-json-remove-recursion-nextstrain.vercel.app/?dataset-json-url=https://nextstrain.org/rsv/a/F@2024-05-27

PR #1474, opened 2024-06-04
feat: warn when standalone ref and tree ref don't match exactly

ivan-aksamentov:
Following conversation in https://github.com/nextstrain/nextclade/pull/1455#issuecomment-2146589157

Let's add a warning if the reference sequence provided any of the possible ways (fasta in dataset files, through CLI argument, Web URL param, or Web "Customization" interface) does not exactly match (as in string comparison) the `.root_sequence.nuc` in Auspice JSON.

The warning is emitted:

 - CLI: into console
 - Web: into dev console, for every WebWorker "thread" (we don't currently have a good mechanism to display warnings int he app, especially the ones coming from Rust)

The warning message is the following. Please suggest improvements (paste a full quote into reply message or feel free to modify in the code).

<details>
<summary>Click to expand</summary>

> Nextclade detected that reference sequence provided does not exactly match reference (root) sequence in Auspice JSON.
>
> This could be due to one of the reasons:
>
> - Nextclade dataset author provided reference sequence and reference tree that are incompatible
> - The reference tree has been constructed incorrectly
> - The reference sequence provided using `--input-ref` CLI argument is not compatible with the reference tree in the dataset
> - The reference tree provided using `--input-tree` CLI argument is not compatible with the reference sequence in the dataset
> - The reference sequence provided using `&input-ref` parameter in Nextclade Web URL is not compatible with the reference tree in the dataset
> - The reference tree provided using `&input-tree` parameter in Nextclade Web URL is not compatible with the reference sequence in the dataset
>
> This warning signals that there is a potential for failures if the mismatch is not intended.

</details>

### DIFF
--- a/packages/nextclade-cli/src/dataset/dataset_download.rs
+++ b/packages/nextclade-cli/src/dataset/dataset_download.rs
@@ -12,7 +12,7 @@ use nextclade::io::fasta::{read_one_fasta, read_one_fasta_str};
 use nextclade::io::file::create_file_or_stdout;
 use nextclade::io::fs::{ensure_dir, has_extension, read_file_to_string};
 use nextclade::run::nextclade_wasm::{NextcladeParams, NextcladeParamsOptional};
-use nextclade::tree::tree::AuspiceTree;
+use nextclade::tree::tree::{check_ref_seq_mismatch, AuspiceTree};
 use nextclade::utils::fs::list_files_recursive;
 use nextclade::utils::option::OptionMapRefFallible;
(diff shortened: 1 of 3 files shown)


ivan-aksamentov:
Should the warning also explain why the match is important and perhaps link the docs? It's kinda getting long. Not sure if the enumeration of all of the cases is also important here. It's basically a list of ways of how you can get ref and tree into the program.

ivan-aksamentov:
Added the same check during index rebuild in the data repo for our own datasets: https://github.com/nextstrain/nextclade_data/pull/206

Issue #1511, opened 2024-07-12
Default threads for webapp are set too high


corneliusroemer:
I recently showed some Nextclade web. We dropped some 100 SC2 sequences and I was surprised how long it took for the first results to show up.

The likely culprit is that the number of threads to use was set at 8 (likely the default for their system, as I don't think they have customized it).

On my own system, I've experimented a little and settled on 3 threads - even though I have 10 cores.

There are good reasons to reduce our default to something like 3 for Nextclade web:

1. Speedup is negligent for more threads
2. Thread setup has a startup cost proportional to the number of threads used
3. Most users of Nextclade web use (only) 10-100 sequences

If even for my routine analysis of 3000 SC2 sequences I don't benefit from more than 3 threads, then the average user definitely does not.

On top, startup delay is more important than time to finish analysis as people can already see partial results even if the analysis takes time.

Overall, I propose we set the default number of threads to use for Web to `min(2, cores)`. Those who use Nextclade for large amounts of sequences can up the number of threads, but there will be few. Overall changing the default will make Nextclade snappier for almost everyone and leave better first impressions.

corneliusroemer:
Curiously, we reduced the threads from 8 to 3 in the past - I wonder how this user got 8 and not 3: https://github.com/nextstrain/nextclade/blob/4d62bb7a615bbb2be6da7e05cbe73d65c502046c/docs/changes/CHANGELOG.old.md#L338

corneliusroemer:
I can reproduce that on safari we set the default number of threads to 8! In contrast to what the code suggests:
https://github.com/nextstrain/nextclade/blob/4d62bb7a615bbb2be6da7e05cbe73d65c502046c/packages/nextclade-web/src/helpers/getNumThreads.ts#L4-L7

This is Nextclade 3.8.1 on Safari on my M1 Mac 14.5
<img width="692" alt="Safari 2024-07-12 10 46 07" src="https://github.com/user-attachments/assets/31691ea3-067e-4272-a898-495e59c4cd7c">

PR #1512, opened 2024-07-12
fix(web): clamp default number of threads to 3

ivan-aksamentov:
Resolves: https://github.com/nextstrain/nextclade/issues/1511

Firefox returns some random number for `navigator?.hardwareConcurrency`. In my case is was `16`, which does not correspond to the avaialble hardware in any way.

Here I additionally clamp the value obtained from `navigator?.hardwareConcurrency` to a range from `DEFAULT_NUM_THREADS` to `MAXIMUM_NUM_THREADS` (currently  from 2 to 3).

This solves it for Firefox, but I don't have a physical mac handy to test Safari. Not sure if testing on Browserstack is accurate - it probably runs on a VM.

### DIFF
--- a/packages/nextclade-web/src/helpers/getNumThreads.ts
+++ b/packages/nextclade-web/src/helpers/getNumThreads.ts
@@ -52,7 +52,7 @@ export function getNumThreads() {
   }
 
   let numThreads = navigator?.hardwareConcurrency ?? DEFAULT_NUM_THREADS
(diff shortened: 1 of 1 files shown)


corneliusroemer (approved):
Nice! The only downside is one can't use more than max threads anymore, but I don't think there's a need to do so, 3 are enough, it doesn't scale anyways higher than that.

I tested on Safari and it works.

Safari 2024-07-12 17 10 52

Issue #1513, opened 2024-07-19
Add coverage per CDS to output


joverlee521:
Motivated by https://github.com/nextstrain/pathogen-repo-guide/issues/50: 

> It seems like a common pattern for sequencing efforts to focus on specific genes instead of the full genome. It would be helpful for ingest to annotate each record's gene coverage to explore the data.

We are currently doing this in what feels like a very "hacky" way of using the Nextclade translations outputs to calculate gene coverage (e.g. [dengue](https://github.com/nextstrain/dengue/blob/029702275f1e72eba07c9339185a2feb8eec942e/ingest/scripts/calculate-gene-converage-from-nextclade-translation.py)). 

Would it possible to directly output gene/CDS coverage from Nextclade?

PR #1514, opened 2024-07-19
feat: calculate cds coverage

ivan-aksamentov:
Resolves: #1513 

 - [x] add a new column/field  "cdsCoverage" to the output files
 - [x] add CDS coverage table in the tooltip of the "Cov."  column in Nextclade Web


rneher:
the algorithm makes sense to me and I think works as intended. 

HIV is a good test case (frame shifts, multi-segment CDS, gaps, insertions, etc):

https://nextclade-git-feat-aa-coverage-nextstrain.vercel.app/?dataset-name=community/neherlab/hiv-1/hxb2&input-fasta=example

Currently it the calculation is `(aligned amino acids + gaps)/reference_length` where numerator and denominator are summed over all segments. Since insertions are not counted, this is always between 0 and 1. Gaps are part of the alignment and I would argue that a protein with a (functional) deletion is as complete at the reference one. So I think the treatment of gaps and insertions is sensible. 

notably, frameshifted regions or `XXXXX` don't count towards the coverage, which makes sense for the intended use case (finding sequences that are worth analyzing). 

If we weren't using "Coverage" for the corresponding number for nucleotides, I would probably argue for "Completeness". 

regarding precision. I don't feel strongly. Digits past `1+log10` of the reference CDS length are not meaningful. But since most genes that are fully covered are just output as 1, this doesn't lead to a huge expansion of the file.

Issue #1525, opened 2024-10-06
Assigned Clade different from Clade of assigned node


taeschbacher-gilead:
I queried a RSVA sequence using the nextclade using the web interface. 

It seems that the sequence was added to a node with clade A.D.3.3. The clade that was assigned to the sequence is A.D.3. Is this an issue or am I not interpreting this correctly?

![rsva_l_nextclade](https://github.com/user-attachments/assets/1bb1d474-4200-4492-8273-83a35c38799d)

This is the fasta file:
[seq.txt](https://github.com/user-attachments/files/17270617/seq.txt)

rneher:
Thanks for raising this, @taeschbacher-gilead !

Nextclade first finds the closest match on the fixed reference tree for each sequence and assigns the clade based on this match. Only after all sequences are assigned to clades, the tree is modified by introducing additional internal nodes to improve the placement of the query sequences. 

The inconsistency you observe is due to the simplistic way nextclade currently labels newly created internal nodes. It currently uses the clade annotation of the child node of the branch that is being split (A.D.3.3 in this case), even though the sibling (your query sequence) is assigned A.D.3. We are working on ways to make the clade assignment of new internal nodes more consistent.

thanks again for pointing this out!
richard

taeschbacher-gilead:
Thank you, Richard, for the fast reply!

PR #1526, opened 2024-10-07
fix: clade mismatch against nearest node

ivan-aksamentov:
Resolves https://github.com/nextstrain/nextclade/issues/1525

After the query node placement is adjusted during the [greedy tree building](https://docs.nextstrain.org/projects/nextclade/en/stable/user/algorithm/03-phylogenetic-placement.html#tree-building), sometimes the branch needs to be split and a new internal node inserted.

Currently we copy the clade of this internal node from the attachment target node. However, this is not always correct and can lead to mismatch between clade of the query node and of the new internal node.

Here I add a voting mechanism (simply a mode) between clades involved: of the parent, target and query nodes.


rneher (approved):
This looks good to me. But this only works for the default clade attribute at the moment, not for the dynamics clade-like attributes.

PR #1527, opened 2024-10-14
feat: Support a NEXTCLADE_EXTRA_CA_CERTS environment variable and --extra-ca-certs option

tsibley:
Updated to [split](https://github.com/nextstrain/nextclade/pull/1527#issuecomment-2412444123) into two parts:

1. https://github.com/nextstrain/nextclade/pull/1529 to use the platform's certification verification, including system-level trust store
2. This PR to add support for configuration of extra CA certificates to trust.

---

_Previously_

See commit messages: 8816fe28692492f839a58b5e2b71022dd63955f8 and 40055050a8f7f621faa99744f4eb93461f08453c.

Motivated by having to write:

> There is currently no way to configure or modify the trust store without modifying the Nextclade source code.

in the recent [CA certificates doc](https://docs.nextstrain.org/en/latest/reference/ca-certificates.html#nextclade-cli) I wrote.

### DIFF
--- a/packages/nextclade-cli/src/io/http_client.rs
+++ b/packages/nextclade-cli/src/io/http_client.rs
@@ -1,6 +1,7 @@
 use clap::{Parser, ValueHint};
 use eyre::{Report, WrapErr};
 use log::info;
+use nextclade::io::file::open_file_or_stdin;
 use nextclade::make_internal_error;
 use nextclade::utils::info::{this_package_name, this_package_version_str};
 use reqwest::blocking::Client;
@@ -10,8 +11,6 @@ use rustls_pemfile;
 use rustls_pki_types::CertificateDer;
 use rustls_platform_verifier::Verifier;
 use std::env;
-use std::io::BufReader;
-use std::fs::File;
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
@@ -128,12 +127,9 @@ impl HttpClient {
 fn extra_ca_certs<'a>() -> Result<impl IntoIterator<Item = CertificateDer<'a>>, Report> {
   match env::var_os("NEXTCLADE_EXTRA_CA_CERTS") {
     Some(filename) => {
-      let file = File::open(filename.clone())
-        .wrap_err_with(|| format!("When opening NEXTCLADE_EXTRA_CA_CERTS file {filename:?}"))?;
+      let mut file = open_file_or_stdin(&Some(filename))?;
 
-      let mut reader = BufReader::new(file);
-
-      let certs = rustls_pemfile::certs(&mut reader)
+      let certs = rustls_pemfile::certs(&mut file)
         .map(|c| c.wrap_err("When parsing an extra CA certificate"))
         .collect::<Result<Vec<_>, Report>>()?;
 
--- a/packages/nextclade-cli/src/io/http_client.rs
+++ b/packages/nextclade-cli/src/io/http_client.rs
@@ -1,12 +1,19 @@
 use clap::{Parser, ValueHint};
-use eyre::Report;
+use eyre::{Report, WrapErr};
 use log::info;
 use nextclade::make_internal_error;
 use nextclade::utils::info::{this_package_name, this_package_version_str};
 use reqwest::blocking::Client;
 use reqwest::{Method, Proxy};
-use rustls_platform_verifier;
+use rustls::ClientConfig;
+use rustls_pemfile;
+use rustls_pki_types::CertificateDer;
+use rustls_platform_verifier::Verifier;
+use std::env;
+use std::io::BufReader;
+use std::fs::File;
 use std::str::FromStr;
+use std::sync::Arc;
 use std::time::Duration;
 use url::Url;
 
@@ -62,8 +69,13 @@ impl HttpClient {
 
     let user_agent = format!("{} {}", this_package_name(), this_package_version_str());
 
+    let tls_config = ClientConfig::builder()
+      .dangerous() // …but the rustls_platform_verifier::Verifier is safe
+      .with_custom_certificate_verifier(Arc::new(Verifier::new_with_extra_roots(extra_ca_certs()?)?))
+      .with_no_client_auth();
+
     let client = client_builder
-      .use_preconfigured_tls(rustls_platform_verifier::tls_config())
+      .use_preconfigured_tls(tls_config)
       .connection_verbose(verbose)
       .connect_timeout(Some(Duration::from_secs(60)))
       .user_agent(user_agent)
@@ -110,5 +122,23 @@ impl HttpClient {
       .bytes()?
       .to_vec();
     Ok(content)
+  }
+}
+
+fn extra_ca_certs() -> Result<impl IntoIterator<Item = CertificateDer<'static>>, Report> {
+  match env::var_os("NEXTCLADE_EXTRA_CA_CERTS") {
+    Some(filename) => {
+      let file = File::open(filename.clone())
+        .wrap_err_with(|| format!("When opening NEXTCLADE_EXTRA_CA_CERTS file {filename:?}"))?;
+
+      let mut reader = BufReader::new(file);
+
+      let certs = rustls_pemfile::certs(&mut reader)
+        .map(|c| c.wrap_err_with(|| "When parsing an extra CA certificate".to_owned()))
+        .collect::<Result<Vec<_>, Report>>()?;
+
+      Ok(certs)
+    }
+    None => Ok(vec![])
   }
 }
--- a/packages/nextclade-cli/src/io/http_client.rs
+++ b/packages/nextclade-cli/src/io/http_client.rs
@@ -134,7 +134,7 @@ fn extra_ca_certs<'a>() -> Result<impl IntoIterator<Item = CertificateDer<'a>>, 
       let mut reader = BufReader::new(file);
 
       let certs = rustls_pemfile::certs(&mut reader)
-        .map(|c| c.wrap_err_with(|| "When parsing an extra CA certificate".to_owned()))
+        .map(|c| c.wrap_err("When parsing an extra CA certificate"))
         .collect::<Result<Vec<_>, Report>>()?;
 
       Ok(certs)
--- a/packages/nextclade-cli/src/io/http_client.rs
+++ b/packages/nextclade-cli/src/io/http_client.rs
@@ -11,6 +11,7 @@ use rustls_pemfile;
 use rustls_pki_types::CertificateDer;
 use rustls_platform_verifier::Verifier;
 use std::env;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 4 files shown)


packages/nextclade-cli/src/io/http_client.rs
```diff
@@ -112,3 +125,26 @@ impl HttpClient {
...
+      let anchors = certs
+        .into_iter()
+        .map(|c| anchor_from_trusted_cert(&c).wrap_err_with(|| "When converting an extra CA certificate to a trust anchor".to_owned()).map(|a| a.to_owned()))
+        .collect::<Result<Vec<_>, Report>>()?;
+
+      Ok(anchors)
+    }
+    None => Ok(vec![])
+  }
+}
```
tsibley, line 150: @ivan-aksamentov I'd appreciate any feedback you have on this code. This isn't the first Rust I've written, but I'm still pretty new to it!

tsibley:
The macOS build is failing because `rustls_platform_verifier::Verifier::new_with_extra_roots()` seems to be available on a platform-specific basis or something? Hmm, I'll dig. This might have been docs.rs misleading me with Linux-specific docs rather than cross-platform/universal docs. (The Linux builds work fine for me and seemingly in CI too.)

The clippy linting fails due to issues in wasm-bindgen, which was [fixed with 0.2.93](https://github.com/rustwasm/wasm-bindgen/blob/0.2.93/CHANGELOG.md#0293). I was hesitant to upgrade wasm-bindgen more than necessary (i.e. beyond 0.2.87 → 0.2.89) because it's a [ZeroVer](https://0ver.org) package without any breakage guarantees and I didn't feel well-placed to evaluate the impact of many of the changes. But I guess if we want Clippy to be happy we gotta?

tsibley:
> The macOS and Windows build is failing because `rustls_platform_verifier::Verifier::new_with_extra_roots()` seems to be available on a platform-specific basis or something?

Ah ha. It is indeed Linux-only as of rustls-platform-verifier 0.3.4 (the latest release) but is being extended to other platforms too: https://github.com/rustls/rustls-platform-verifier/issues/58. Implementations for macOS and Windows are landed but not yet released.

tsibley:
If we don't want to wait for a new release of rustls-platform-verifier (nor build from GitHub), we can take just the first commit in this PR (8816fe2) and land the second commit later.

packages/nextclade-cli/src/io/http_client.rs
```diff
@@ -135,12 +135,12 @@ fn extra_ca_certs<'a>() -> Result<impl IntoIterator<Item = TrustAnchor<'a>>, Rep
       let mut reader = BufReader::new(file);
 
       let certs = rustls_pemfile::certs(&mut reader)
-        .map(|c| c.wrap_err_with(|| "When parsing an extra CA certificate".to_owned()))
+        .map(|c| c.wrap_err("When parsing an extra CA certificate"))
```
tsibley: Ohhh, this makes lots of sense!

packages/nextclade-cli/src/io/http_client.rs
tsibley: Thanks!

packages/nextclade-cli/src/io/http_client.rs
```diff
@@ -61,7 +69,13 @@ impl HttpClient {
 
     let user_agent = format!("{} {}", this_package_name(), this_package_version_str());
 
+    let tls_config = ClientConfig::builder()
+      .dangerous() // …but the rustls_platform_verifier::Verifier is safe
+      .with_custom_certificate_verifier(Arc::new(Verifier::new_with_extra_roots(extra_ca_certs()?)))
+      .with_no_client_auth();
```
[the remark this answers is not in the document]
tsibley: It does disable client-sent certificates, which I don't think I've seen any of our users need. Browsers _can_ send client certificates, though in most cases that's not used. I don't know if it's strictly necessary for the `rustls-platform-verifiers` crate. It comes from the [example in rustls-platform-verifier's doc](https://docs.rs/rustls-platform-verifier/0.3.4/rustls_platform_verifier/index.html#installation-and-setup), and I noted it was also in the [implementation of `rustls_platform_verifier::tls_config()`](https://docs.rs/rustls-platform-verifier/0.3.4/src/rustls_platform_verifier/lib.rs.html#61-66), which is the "out of the box (no extra certs)" usage of that crate.

tsibley:
> Let's hope that their planned implementations will work in cross environments too.

As far as I saw, implementing `Verifier::new_with_extra_roots()` for the other platforms we care about doesn't mean new libraries or new platform-specific integrations, but just some more Rust code to call existing libraries and use existing integrations.  Point is, if you can cross-build `rustls-platform-verifier` 0.3.4 now, then I think you'll be fine.

> For the time being we could extract the Linux code to a function and slap conditional compilation to it. What do you think?

I guess, but I'd be surprised if `NEXTCLADE_EXTRA_CA_CERTS` only worked on Linux?

I'd probably just drop it for now, leaving the new system-level trust store support in place, as that's a big enough win by itself.  Linux users will still have other ways of adding extra CA certs (e.g. via OpenSSL's `SSL_CERT_FILE`).

> Does it make sense to also add a CLI flag, just in case? #1528

Sure.

> This also allows us to make the feature more visible and document it right in the help screen.

+1 for doc there.  I was gonna ask where this should got doc'd.

Relatedly: should I add tests for this (or at least the env var)?  Where would those go?  The few Cram tests I saw look unused, but that's the kind of thing I'd reach for with this particular feature.

> What is the usual convention: flag has priority over env var or other way around?

Yes, usually argument overrides env var, as its more specific to a single invocation.  Env vars can be thought of overriding or setting the defaults.

>  Should we also check for `REQUESTS_CA_BUNDLE` (mentioned in Nextstrain docs) and perhaps `CURL_CA_BUNDLE`?
>
> Explicit Nextclade-specific `NEXTCLADE_EXTRA_CA_CERTS` var makes sense for direct users, but from Nextstrain user's perpective where Nextclade might not even appear directly, having to track a whole zoo of env vars might not be be perfect.
>
> Should we try to cover the most widespread ones? Or is it looking for troubles?

Agreed a zoo of env vars is not perfect, but that's reality.  I think it's looking for troubles.  It's surprising to me to have software which goes looking for env vars expected by other software (e.g. `REQUESTS_CA_BUNDLE`).  The best way to avoid a zoo is to add certs to the system trust store and expect/configure everything to use that.

If we want to insulate Nextstrain users specifically (though I don't think we need to here), I'd instead advocate for a specific `NEXTSTRAIN_EXTRA_CA_CERTS` and adapting _all_ our software to DTRT with it.

> > So any platform-specifics might be pretty difficult to handle, even if they are supported. Let's hope that their planned implementations will work in cross environments too.
> 
> For the time being we could extract the Linux code to a function and slap conditional compilation to it. What do you think?
> 
> 90% of CLI downloads are for Linux. I guess there might be some Windows and Mac users, but having this at least for Linux is already a big win, I think!

> When considering a CLI flag in #1528, what are the possible good names for the flag to harmonize it across Nextstrain and perhaps also with `curl`? Looks like `curl` accepts `--with-ca-path` which accepts a directory path.

Curl is basing those on OpenSSL's conventions.  When given a directory, OpenSSL in particular requires a specific content-addressed filename scheme that I don't think we want to explicitly support.  And when Nextclade is using OpenSSL with this branch, then the normal `SSL_CERT_DIR` will still be available.

In general, I wouldn't go searching for functionality we _could_ add here but focus on the minimum needed to solve concrete problems (e.g. no way to configure CA certs) and meet basic expectations (e.g. system CAs are used).

> It seems that there could be multiple certs in the file, and when the libraries process it, some certs could fail. Currently the entire program fails if one cert fails to load.
>
> We could consider skipping failing certs, if that makes sense and is safe, though I cannot immediately evaluate either.
>
> One disadvantage of silently failing certs is that it will be unclear for the user that e.g. a single cert in a file fails. It will probably only manifest in connection errors later. To remediate, we could emit a warning for each failing cert.

I intentionally chose to fail hard early if anything doesn't succeed in the extra cert adding steps.  I don't think there's any advantage to allowing invalid certs in the explicitly-configured file.  Especially with security stuff, I'd rather know there's a problem or deviation from what I expect earlier than later.

> We could also consider to bubble the check for presence for env var (and/or cli arg) up and only run the new code path conditionally if the CA certs are provided. Just to be extra safe that this does not affect the old code path, which most users use (especially considering the 0.x nature of all of these libs).

Hmm.  I think you're conflating two things here.  Using the new code path only when extra CA certs are provided sort of defeats the point.  The extra CA certs are merely a convenience.  The real useful change is using the system CA trust store, which should be the default and expected behaviour for any software (and it seems that reqwest is moving that way eventually as well).  It means that for many correctly-configured systems there would be no additional configuration required for Nextclade to Just Work.

If you _didn't_ want to use the system CA trust store and instead _only_ support configuration of extra CA certs explicitly, then that can be done without `rustls-platform-verifier` entirely.  But again, that'd be throwing away the change that I think is most useful.

packages/nextclade-cli/src/io/http_client.rs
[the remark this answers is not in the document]
tsibley: It's also what reqwest does by default when it constructs the TLS config (i.e. instead of us providing one as we do in this PR):

https://docs.rs/reqwest/latest/src/reqwest/async_impl/client.rs.html#611-615

as we don't construct a reqwest client with an [`identity` field](https://docs.rs/reqwest/latest/reqwest/blocking/struct.ClientBuilder.html#method.identity).

tsibley:
> But I would not test an env var, as it's just a single call to the std library. It will probably take too much time to test anything else in the PR.

Ah, I meant test that setting the env var correctly gets a cert into the trust store and that it allows the TLS connection to succeed. An end-to-end integration test, exactly like what I've been doing manually to verify the functionality on my local machine. The actual specific test set up for this on Linux and macOS would be very straightforward for me. Windows would be a little more unknown, but I think I know how I'd do it. But if there's existing scaffolding for testing the final `nextclade` binaries isn't there… then yeah, maybe too much work to have to build that out first too? Also, if you're cross compiling to macOS and Windows, is there native test infra for those platforms available?

>  Sounds like a plan! […] Correct?

Correct! I'll do it now on a new PR and leave this one with the extended conversation for revisiting.

tsibley:
> I'll do it now on a new PR and leave this one with the extended conversation for revisiting.

Done. I've also gone ahead and reworked this PR to use the unreleased rustls-platform-verifier changes from GitHub (pinned to a specific revision) so that we can a) verify it works as expected and b) we're ready for when those changes are released.

I've verified the functionality still works as expected, at least on x86_64-unknown-linux-gnu.

tsibley:
Main commit from me now reads:

```
feat: Support a NEXTCLADE_EXTRA_CA_CERTS environment variable

Allows adding additional CA certificates to the trust store specifically
for Nextclade, for when modifying the system's trust store isn't
desirable/possible.  Works across all platforms.

Currently requires using landed unreleased changes to
rustls-platform-verifier so that macOS and Windows have the needed
Verifier::new_with_extra_certs() API.¹  That API also has signature
changes proposed² which, while unmerged, seem likely to land, so use
those in anticipation.  They actually bring the signature back closer to
the latest release (0.3.4).

¹ <https://github.com/rustls/rustls-platform-verifier/issues/58>
² <https://github.com/rustls/rustls-platform-verifier/pull/145>

Related-to: <https://github.com/nextstrain/nextclade/pull/1529>
Related-to: <https://github.com/nextstrain/nextclade/issues/726>

```

tsibley:
Closing this as I expect it to be superseded by https://github.com/nextstrain/nextclade/pull/1536.

PR #1528, opened 2024-10-14
feat(cli): add --extra-ca-certs

ivan-aksamentov:
Followup of https://github.com/nextstrain/nextclade/pull/1527

Adds a CLI arg with the same meaning as `NEXTCLADE_EXTRA_CA_CERTS` env var - to provide a path to extra CA certs. Perhaps some users might prefer a CLI arg. It was easy, so I thought why not?

### DIFF
--- a/packages/nextclade-cli/src/io/http_client.rs
+++ b/packages/nextclade-cli/src/io/http_client.rs
@@ -11,6 +11,7 @@ use rustls_pemfile;
 use rustls_pki_types::TrustAnchor;
 use rustls_platform_verifier::Verifier;
 use std::env;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


packages/nextclade-cli/src/io/http_client.rs
```diff
@@ -34,6 +35,13 @@ pub struct ProxyConfig {
   #[clap(long)]
   #[clap(value_hint = ValueHint::Other)]
   pub proxy_pass: Option<String>,
+
+  /// Path to extra CA certificates
+  ///
+  /// You can also provide path to CA certificates in environment variable `NEXTCLADE_EXTRA_CA_CERTS`. The argument takes precedence over environment variable if both are provided.
+  #[clap(long)]
+  #[clap(value_hint = ValueHint::Other)]
+  pub extra_ca_certs: Option<PathBuf>,
```
tsibley, line 44: This is in `ProxyConfig`, but it's not really a proxy configuration item. It may be used often with proxies, but it won't _only_ be used with proxies.

tsibley: I think `--extra-ca-certs` is the right name given `NEXTCLADE_EXTRA_CA_CERTS`. We could switch from "CA" to "root" if we wanted, but "CA certs" is a pretty common term-of-art.

tsibley: (I wouldn't say just `--ca-certs` because that implies replacing _all_ the default CA certs too and is how that option is used by other software.)

Issue #726, opened 2022-02-09
Error using `nextclade dataset get` behind proxy: CURLE_PEER_FAILED_VERIFICATION: SSL peer certificate or SSH remote key was not OK


EricFournier3:
nextclade version 1.10.2 on centos-release-7-4.1708.el7.centos.x86_64

Hi, 

when I execute the following command, 
`nextclade dataset get --name sars-cov-2 --output-dir /home/foueri01@inspq.qc.ca/temp/20220209/TESTNEXTLADE`
I get the following error
**[ERROR] Nextclade: When fetching a file "https://data.clades.nextstrain.org/index.json": CURLE_PEER_FAILED_VERIFICATION: SSL peer certificate or SSH remote key was not OK**
I got the same error when I append **insecure** in my ~/.curlrc

Actually, our process run under a proxy with ssl full inspection. The non-recognized certificate is on our side. Is there a way like curl to activate a switch to ignore the certificate ?

Can you please help me with this
Best,
Eric

corneliusroemer:
I'm not very familiar with proxy servers and how they might interact with curl here. Have you tried setting the following flags?

```bash
CURLOPT_SSL_VERIFYPEER=0
CURLOPT_PROXY_SSL_VERIFYPEER=0
CURLOPT_SSL_VERIFYHOST=0
CURLOPT_PROXY_SSL_VERIFYHOST=0
nextclade get ...
```

I found them here and through StackOverflow: https://www.php.net/manual/en/function.curl-setopt.php

@ivan-aksamentov may know better things to try.

In any case, you can download in any way you like, using `wget` or `curl`, that is without using the convenience command `nextclade dataset get`. The implementation could however change at some point in the future, so it's best to sort out `nextclade dataset get`. But for now to be able to use datasets go ahead as follows.

All information is contained in this json: https://data.clades.nextstrain.org/index.json

The latest dataset has the field `latest: true`. The URL to the zip folder is:
`wget https://data.clades.nextstrain.org/datasets/sars-cov-2/references/MN908947/versions/2022-02-07T12:00:00Z/zip-bundle/nextclade_dataset_sars-cov-2_MN908947_2022-02-07T12:00:00Z.zip
`

Unzip that into a folder and you're ready to go. You may still run into proxy issues, but at least the problem is simplified to "download a zip file" and you don't need to worry about how Nextclade invokes curl.

corneliusroemer:
[...]

PR #1529, opened 2024-10-15
feat: Switch to platform-specific TLS/SSL certificate verification

tsibley:
Use rustls-platform-verifier to verify certificates with the OS/platform's standard methods, including the system's standard CA trust store.  See rustls-platform-verifier's documentation for details on each platform.¹

Previously, the system's trust store was ignored in favor of a baked in and unconfigurable trust store provided by webpki-roots.

Updates reqwest as necessary to ensure a single compatible rustls version remains in use.  Updates wasm-bindgen as necessary for the newer reqwest.

¹ <https://github.com/rustls/rustls-platform-verifier/blob/v/0.3.4/README.md>

Resolves: <https://github.com/nextstrain/nextclade/issues/726>
Related-to: <https://github.com/nextstrain/nextclade/pull/1527>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3388,11 +3388,12 @@ checksum = "9c8d87e72b64a3b4db28d11ce29237c246188f4f51057d65a7eab63b7987e423"
 
 [[package]]
 name = "wasm-bindgen"
-version = "0.2.89"
+version = "0.2.93"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0ed0d4f68a3015cc185aff4db9506a015f4b96f95303897bfa23f846db54064e"
+checksum = "a82edfc16a6c469f5f44dc7b571814045d60404b55a0ee849f9bcfa2e63dd9b5"
 dependencies = [
  "cfg-if",
+ "once_cell",
  "serde",
  "serde_json",
  "wasm-bindgen-macro",
@@ -3400,9 +3401,9 @@ dependencies = [
 
 [[package]]
 name = "wasm-bindgen-backend"
-version = "0.2.89"
+version = "0.2.93"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1b56f625e64f3a1084ded111c4d5f477df9f8c92df113852fa5a374dbda78826"
+checksum = "9de396da306523044d3302746f1208fa71d7532227f15e347e2d93e4145dd77b"
 dependencies = [
  "bumpalo",
  "log",
@@ -3427,9 +3428,9 @@ dependencies = [
 
 [[package]]
 name = "wasm-bindgen-macro"
-version = "0.2.89"
+version = "0.2.93"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0162dbf37223cd2afce98f3d0785506dcb8d266223983e4b5b525859e6e182b2"
+checksum = "585c4c91a46b072c92e908d99cb1dcdf95c5218eeb6f3bf1efa991ee7a68cccf"
 dependencies = [
  "quote",
  "wasm-bindgen-macro-support",
@@ -3437,9 +3438,9 @@ dependencies = [
 
 [[package]]
 name = "wasm-bindgen-macro-support"
-version = "0.2.89"
+version = "0.2.93"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f0eb82fcb7930ae6219a7ecfd55b217f5f0893484b7a13022ebb2b2bf20b5283"
+checksum = "afc340c74d9005395cf9dd098506f7f44e38f2b4a21c6aaacf9a105ea5e1e836"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -3450,9 +3451,9 @@ dependencies = [
 
 [[package]]
 name = "wasm-bindgen-shared"
-version = "0.2.89"
+version = "0.2.93"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7ab9b36309365056cd639da3134bf87fa8f3d86008abf99e612384a6eecd459f"
+checksum = "c62a0a307cb4a311d3a07867860911ca130c3494e8c2719593806c08bc5d0484"
 
 [[package]]
 name = "wasm-logger"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -108,7 +108,7 @@ version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5ca11d4be1bab0c8bc8734a9aa7bf4ee8316d462a08c6ac5052f888fef5b494b"
 dependencies = [
- "windows-sys",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -118,7 +118,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "58f54d10c6dfa51283a066ceab3ec1ab78d13fae00aa49243a45e4571fb79dfd"
 dependencies = [
  "anstyle",
- "windows-sys",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -213,9 +213,9 @@ dependencies = [
 
 [[package]]
 name = "base64"
-version = "0.21.2"
+version = "0.22.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "604178f6c5c21f02dc555784810edfb88d34ac2c73b2eae109655649ee73ce3d"
+checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "base64ct"
@@ -299,9 +299,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.3.3"
+version = "2.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "630be753d4e58660abd17930c71b647fe46c27ea6b63cc59e1e3851406972e42"
+checksum = "b048fb63fd8b5923fc5aa7b340d8e156aec7ec02f0c78fa8a6ddc2613f6f71de"
 
 [[package]]
 name = "block-buffer"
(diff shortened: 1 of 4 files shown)


tsibley:
Looks like the builds went well in CI for all targets, but a few checks failed because there were [no binaries uploaded](https://github.com/nextstrain/nextclade/actions/runs/11339823592/job/31535330309?pr=1529#step:17:14) and so [none to download for those checks](https://github.com/nextstrain/nextclade/actions/runs/11339823592/job/31535650822?pr=1529#step:6:10). That's because the binaries are written to `.build/…` but `.out/…` is what's uploaded as a workflow artifact for subsequent steps, and the conditions for `./docker/dev` to copy from `.build/` → `.out/`

https://github.com/nextstrain/nextclade/blob/9b2d9ddbfc42af1ddd646fb0c00f3ba1da1431b1/docker/dev#L573-L580

weren't met. Huh. This seems intentional (?) but I admit I don't get why.

tsibley:
> I believe that the bash checks should be met: running with `br` (`build-release`) means both BUILD=1 and RELEASE is set, and then CROSS should be set - the env var is provided explicitly when calling the script

Oh, you're totally right. I misread. It seems as if it's either 1) never getting to the `cp`, as `cp` would complain if anything was missing or 2) it's getting to the `cp` just fine, but for some other reason the upload-artifact step isn't seeing `.out/`.

And writing that out, it just clicked. `actions/upload-artifact@v4` made some pretty big breaking changes—I recall from when I upgraded to v4 in other repos—and indeed, [here's](https://github.com/actions/upload-artifact/blob/184d73b71b93c222403b2e7f1ffebe4508014249/README.md#breaking-changes) the one that's biting us now:

> With v4.4 and later, hidden files are excluded by default.

We need to set:

```yaml
include-hidden-files: true
```

in its params. Maybe also `if-no-files-found: error` instead of the default of only warning, since it seems like binaries should always be present.

tsibley:
> If you noticed potential breakage sources or have any ideas regarding improving the (admittedly horrible) dev script, the container and CI setup, I appreciate your advice :)

The dev setup is pretty… complicated. I can get some of it to work for me, but other parts are a little broken in various ways due to little details that don't line up. The sort of thing that accumulates when only a couple people regularly run a program and then someone new tries to run it.

- Docker images are pulled, but then `docker buildx build` is still run immediately afterwards. This seems like it's not necessary? Maybe it's a quick-enough no-op in that case, but I couldn't tell because I ran into `ERROR: cache export feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")`. I know what's going on and how to fix this… but I realized I didn't have to since I didn't need to build.

- `./docker/dev build` works for me now, but `CROSS=x86_64-unknown-linux-gnu ./docker/dev build` does not: `/usr/bin/time: cannot run cargo: Permission denied`. My uid is 1000 but it seems like the cross container's files are owned by and only visible to 1001. My host is x86_64-unknown-linux-gnu.

I don't have the energy to fix 'em now, and I haven't really had to run `./docker/dev` to hack on things yet, so… *shrug*

tsibley:
Presumably if I fix the buildx builder instance issue and force a rebuild with `./docker/dev docker-image-build` (because it won't be rebuilt now that they've been pulled), then the cargo permission denied issue will go away. But also presumably a better set umask in the images would make some of the UID stuff unnecessary.

tsibley:
[Fixing CI](https://github.com/nextstrain/nextclade/pull/1533) revealed another [CI issue](https://github.com/nextstrain/nextclade/actions/runs/11349833151/job/31567441936?pr=1533#step:5:40) (and maybe larger issue?) with this PR: systems without a trust store error, e.g. like the bare bones Linux containers without ca-certificates installed that we use in `./tests/test-linux-distros`. We could address this with 1) default certs from [webpki-roots](https://crates.io/crates/webpki-roots) in code or 2) in the CI setup by ensuring `ca-certificates` is installed. The "right" solution depends on if such an environment without a system trust store matters for actual use cases. I guess in other user-built containers it might already be relied upon (even without realizing it)?

tsibley:
> As mentioned in the comment in the dev script, none of this is required. It's just basically my personal giant set of bash aliases :)
> […]
> But then I dragged it into CI, and now there is no easy way back.

Yeah, that's why I ignored it. :-)

But then I wanted to test what CI was testing…

> Could it be a problem with docker version?

Ah, no, but I do know what the issue is (as I said): there's an incorrect assumption that the default buildx builder driver is `docker-container` not `docker`. (We [handle this in nextstrain/docker-base by not assuming that](https://github.com/nextstrain/docker-base/blob/0acdc79904cea7d9b6195ac3ae3028ef588bef1d/devel/build#L46-L56).) I might address it so I can locally 1) build a proper binary against an old-enough glibc and 2) run `tests/test-linux-distros`.

> I ship [debian](https://github.com/nextstrain/nextclade/blob/master/docker/docker-prod-debian.dockerfile), [alpine](https://github.com/nextstrain/nextclade/blob/master/docker/docker-prod-alpine.dockerfile) and [scratch](https://github.com/nextstrain/nextclade/blob/master/docker/docker-prod-scratch.dockerfile) images, so that might need to be adjusted somehow.

I expect Debian image would be fine (ca-certificates is installed), but that the Alpine and scratch images would not be.

> Could there be also problems in Conda envs? I am not sure what they use.

Conda envs basically always have ca-certificates installed on Linux, and on other systems they'll use the platform.

I don't really want to say "user responsibility" as it feels unfriendly and is essentially a breaking change. I'll address this in additional code to fall back to webpki-roots (which is what was always used previously).

tsibley:
> Last time I tried rootless docker a couple of years ago it was a major pain and I failed to even set it up on my system.

You might look into [Podman](https://podman.io/), which has a much simpler rootless configuration due to its design not using a daemon (docker + dockerd) but forking+execing directly.

> Now you mentioned umask. If you know a way to set things up such that UID/GID are not baked-into the image, can you teach me please? Does it have limitations compared to setting UID/GID?

(In case it's not clear, the umask I'm talking about is `umask` ([man page](https://www.man7.org/linux/man-pages/man1/umask.1p.html), [Wikipedia](https://en.wikipedia.org/wiki/Umask)). I used "umask" as a shorthand for "default file permissions".)

I mentioned it because the uid/gid problems I ran into only seemed to arise in the first place because files/directories were set to be readable/executable only by the owner and no one else.  That might because of the umask setting at image build time, or it might be because something explicitly set them that way.  Regardless, in a container context with a single user, it rarely makes the most sense to have owner-only perms, as then the invoking uid (`--user=…`) and baked-in uid (`USER` in `Dockerfile`) have to match.  The problem I ran into was because they didn't: my uid was 1000 and the baked-in uid was 1001 (because I used a pulled image directly instead of rebaking my uid into it).  If the files/dirs were readable/executable by everyone, then this mismatch wouldn't have mattered.

My rules of thumb are to:

1. Make sure that the `Dockerfile` produces images with world-readable/executable (and writable, if applicable) permissions.  `umask` may be a part of this, or you may choose/need to achieve it other ways.  If you want, you can also use a non-root default user, but for tool images like this, that's not strictly necessary if you'll also always be doing (2) below.  (For service images like, say, Pg, it's more necessary as you'd less commonly invoke them with `--user`.)
2. Invoke such images with `--user=$(id -u):$(id -g)` to ensure that any _new_ files/dirs created on mounts shared with the host remain owned by the current user/group.

> I am not very familiar with the new builders stuff, neither I can repro it on my system or on CI, so I appreciate if you look at it. No rush though.

https://github.com/nextstrain/nextclade/pull/1534

> My main question is if it was working for me on different machines over the years and on CI, why it does not work for you? Do you have a different local docker configuration that prevents the docker-container driver assumed by my script from working?

Yeah.  My default buildx builder uses the [`docker` driver](https://docs.docker.com/build/builders/drivers/docker/), not the [`docker-container` driver](https://docs.docker.com/build/builders/drivers/docker-container/).  This is the standard default, so you must have switched yours at some point long ago.

For reference, these are the builders I have right now:

```console
$ docker buildx ls
NAME/NODE             DRIVER/ENDPOINT             STATUS  BUILDKIT             PLATFORMS
nextclade-builder     docker-container                                         
  nextclade-builder0  unix:///var/run/docker.sock stopped                      
nextstrain-builder    docker-container                                         
  nextstrain-builder0 unix:///var/run/docker.sock stopped                      
default *             docker                                                   
  default             default                     running v0.11.7+d3e6c1360f6e linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/386
```

The output indicates the currently selected builder instance (`default` for me) with a `*`.

> We lived past 10 years without it all and now docker devs felt compelled to add this new complexities. I guess that there is a good reason. I need to learn more about it sometimes to avoid falling into these traps.

I hear ya!  I only know about this shit because we dealt with it for nextstrain/docker-base.

tsibley:
> I'll address this in additional code to fall back to webpki-roots (which is what was always used previously).

Reasonable options I see for getting out of the hole I dug (lolsob):

 1. Use platform certs + built-in webpki-roots certs on all platforms…

     1. …using `rustls_platform_verifier::Verifier::new_with_extra_roots()`, similar to what https://github.com/nextstrain/nextclade/pull/1527 does for `NEXTCLADE_EXTRA_CA_CERTS`.

        This relies on currently-unreleased code, so we either accept that and move forward or revert this PR's merge to master and wait until the dep is released to switch to `rustls-platform-verifier`.  Frankly, the former seems ok to me: merged-but-not-released seems very similar to merged-and-released for this package.

     2. …by enabling both the `rustls-tls-native-roots` and `rustls-tls-webpki-roots` features of `reqwest`.  Currently we only enable the latter (via `rustls` feature alias), but we can enable both at once.  This would use [`rustls-native-certs`](https://docs.rs/crate/rustls-native-certs) to acquire the platform certs, if any.  It's better than no platform certs but not quite as good as using the platform's own cert verification methods (via `rustls-platform-verifier`).

        We could also support `NEXTCLADE_EXTRA_CA_CERTS` in this implementation by calling [`reqwest::blocking::ClientBuilder::add_root_certificate()`](https://docs.rs/reqwest/0.12.8/reqwest/blocking/struct.ClientBuilder.html#method.add_root_certificate) repeatedly.

 2. Only add in certs from webpki-roots for Linux, where `rustls_platform_verifier::Verifier::new_with_extra_roots()` is available in released code.  Other platforms _should_ always have a system trust store with certs in it, so _shouldn't_ need webpki-roots.

    When https://github.com/nextstrain/nextclade/pull/1527 lands, still don't add in webpki-roots for non-Linux platforms.

Thoughts?  Mine are:

(1) is more like the Nextclade status quo before any of my work here: it just adds certs from the platform alongside previously used certs from webpki-roots.  (1)(i) is maybe the least change from the status quo.

(1)(i), especially if we accept the unreleased code, is doubling down on switching to `rustls-platform-verifier`.  (1)(ii) is walking it back and changing tack; we could still switch to it later (maybe when/if `reqwest` adopts it, as folks hope?) but maybe we won't.

(1)(i) and (2) preserve `rustls-platform-verifier`'s full use of the platform's own verification methods beyond CA certs themselves, which is nice.  Maybe not really something we need to work hard for, but at this point most of the work is done.

(2) is more "pure" in that it doesn't mix sources of certs unnecessarily.  (1) is maybe more "consistent" though.

Having spent way more time on this now than I bargained for and really want to, I'm inclined to go with (1b): ditch `rustls-platform-verifier`, go with something simpler and less different than what we have now, still solve CA cert problems now not later.

Issue #1532, opened 2024-10-15
CI is broken


ivan-aksamentov:
First noticed by @tsibley in https://github.com/nextstrain/nextclade/pull/1529#issuecomment-2412960515

I tried to make empty commits stemming off [master](https://github.com/nextstrain/nextclade/pull/1530) and off [release](https://github.com/nextstrain/nextclade/pull/1531) branches, and they both have the same result - the binaries seems to be compiling successfully, but nothing gets copied to the `.out` dir. This is strange because the release branch definitely got built and released last time. I don't remember changing anything CI-related in a while.

If I try the same things locally, the

```bash
CROSS=x86_64-unknown-linux-gnu ./docker/dev br
```
produces `.out/nextclade-x86_64-unknown-linux-gnu` as expected.

and 

```bash
./docker/cross
```
produces binaries for all platforms as expected:

![s](https://github.com/user-attachments/assets/9da4fd58-d57b-4fbc-96f5-0b1f38518629)

I might need to clear docker images and retry.

I appreciate any input.

PR #1533, opened 2024-10-15
fix(ci): Upload all CLI build outputs

tsibley:
Include hidden files since our sole path pattern is a hidden "file". Apparently this counts even though the files the pattern matches aren't hidden.

GitHub made this big breaking change of excluding hidden files without bumping the major version number of actions/upload-artifact.¹  It, rightfully so, caused a ruckus at the time and continues to be a gift that keeps on giving.

The only other actions/upload-artifacts usage, in
.github/workflows/web.yml, seemingly isn't affected even though the path contains a hidden directory.  Perhaps because in that case there are a few more directory components after it?

¹ <https://github.com/actions/upload-artifact/releases/tag/v4.4.0>

Resolves: <https://github.com/nextstrain/nextclade/issues/1532>

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -180,6 +180,7 @@ jobs:
         with:
           name: out-${{ matrix.arch }}
           path: ./.out/*
+          include-hidden-files: true
 
 
   run-unit-tests:


tsibley:
This fixes the CI issue… and reveals [another CI issue](https://github.com/nextstrain/nextclade/actions/runs/11349833151/job/31567441936?pr=1533#step:5:40) (and maybe larger issue?) with the merge of #1529: systems without a trust store (like a bare bones Linux container without `ca-certificates` installed) error. We could address that in code, or in the CI setup, depending on if such an environment matters for actual use cases.

PR #1540, opened 2024-10-18
feat: relabel aminoacids when attaching node 

ivan-aksamentov:
Followup of: https://github.com/nextstrain/nextclade/pull/1526

Additionally to moving clade let's also move "aa" label


rneher:
I believe the `aa` are handled already: 

https://github.com/nextstrain/nextclade/blob/8e864cafae904cf92e33f4b9dcc141b01abcd485/packages/nextclade/src/tree/tree_builder.rs#L375

https://github.com/nextstrain/nextclade/blob/8e864cafae904cf92e33f4b9dcc141b01abcd485/packages/nextclade/src/tree/tree_builder.rs#L424

PR #1544, opened 2024-11-01
feat(web): add link to view reference tree on nextstrain org

ivan-aksamentov:


rneher:
I like it. I think we can merge this!

PR #1555, opened 2024-11-20
Adjust Nextclade to Auspice 2.61+

ivan-aksamentov:
Fixes build https://github.com/nextstrain/nextclade/actions/runs/11903714964/job/33171209422?pr=1553#step:18:146

This attempts to adjust Nextclade to changes introduced in Auspice 2.61.0 and above.

This involves a few moved files and making sure ts files are included into monkey patching.

Sadly, entropy panel permanently shows "data loading" after any interaction with the tree (e.g. clicking on a branch or using filters)

I made an isolated `/tree-debug` page to reproduce and investigate the issue further: https://github.com/nextstrain/nextclade/blob/1b7917a62ca34e34af7780eef611eaa717794b69/packages/nextclade-web/src/pages/tree-debug.tsx#L21-L34


jameshadfield:
> Sadly, entropy panel permanently shoes "data loading" after any interaction with the tree (e.g. clicking on a branch or using filters)

At a guess this is due to [this conditional](https://github.com/nextstrain/auspice/blob/f9d7ec75da44b5acbf30f7b5c361d3cd46760f7b/src/actions/entropy.js#L17-L23) where we don't update the data behind the entropy panel if the panel's not rendered or off-screen. I'd guess [the intersection observer](https://github.com/nextstrain/auspice/blob/f9d7ec75da44b5acbf30f7b5c361d3cd46760f7b/src/components/entropy/index.js#L199-L222) we use to know if it's off-screen isn't working as expected within Nextclade. But it may also be the former if Nextclade doesn't use `<reduxState>.controls.panelsToDisplay` like vanilla Auspice does.

jameshadfield:
> Alternatively, do you see an easy way to opt out of this optimization and render the entropy always, as before?

You could change the source code to modify the conditional to `if (false) {`, i.e. [within monkeyPatch.ts](https://github.com/nextstrain/nextclade/blob/c95471b4faa9121feea6102a333672ffcca74a06/packages/nextclade-web/tools/monkeyPatch.ts). It's been a long time since I properly looked at how Nextclade uses Auspice, but [skimming through things now](https://github.com/search?q=repo%3Anextstrain%2Fnextclade+path%3A%2F%5Epackages%5C%2Fnextclade-web%5C%2Fsrc%5C%2F%2F+auspice&type=code&p=1) makes me think you're using a very cut-down version of the (redux) controls state and adding `panelsToDisplay: ['tree']` may be enough.

PR #1569, opened 2025-02-21
feat: implement alignment param presets

ivan-aksamentov:
Define some hard-coded presets of alignment parameters additionally to the default one.

The `--alignment-preset` arg allows users to toggle between presets.

Example:

```bash
nextclade run --dataset-name=sars-cov-2 -O tmp/sars-cov-2 sequences.fasta --alignment-preset=high-diversity --verbosity=info
```


corneliusroemer:
Nice stuff! @anna-parker you might be interested in this one potentially 😀

rneher:
I think this is pretty useful. There no universally optimal parameters, but the one in place now were tested for HIV-1 and for aligning RSV-A to RSV-B. Short sequences were tested with RSV fragments >30 bp. This all worked well.

Issue #1590, opened 2025-03-19
Crash when running with dengue@2025-03-03


ivan-aksamentov:
Nextclade CLI and Web crash on an assertion when running with a particular dengue auspice json (auspice dataset as nextclade dataset) from 2025-03-03.

Reported here: https://github.com/nextstrain/dengue/issues/97#issuecomment-2736302376

CLI repro for debugging:

```bash
docker pull nextstrain/nextclade:latest

docker run -it --rm nextstrain/nextclade:latest bash

# Download auspice dataset JSON
curl -o dengue_all_genome.json -fsSL -H 'Accept: application/vnd.nextstrain.dataset.main+json;q=1, application/json;q=0.9, text/plain;q=0.8, */*;q=0.1' 'https://next.nextstrain.org/dengue/all/genome@2025-03-03'

# Download nextclade dataset (for example data)
nextclade dataset get --name=nextstrain/dengue/all --output-dir=data_dev/nextstrain/dengue/all

# Run nextclade
nextclade run --input-dataset 'dengue_all_genome.json' -O 'tmp/dengue-crash' 'data_dev/nextstrain/dengue/all/sequences.fasta'

The application panicked (crashed).
Message:  assertion `left == right` failed
  left: 10649
 right: 10648
Location: packages/nextclade/src/analyze/nuc_changes.rs:18

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
Aborted (core dumped)
```

Can be reproduced in web with https://clades.nextstrain.org/?dataset-json-url=https://next.nextstrain.org/dengue/all/genome@2025-03-03 and example data for `dengue/all`

PR #1591, opened 2025-03-20
feat: forbid gaps in reference sequence

ivan-aksamentov:
Resolves https://github.com/nextstrain/nextclade/issues/1590

We discovered, that the problem was due to gaps in the reference sequence (position 2984).

This PR disallows gaps in the reference sequence. Nextclade will stop with a descriptive error message instead of crashing now:

```
Error:
   0: When validating reference sequence 'reference'
   1: Sequence contains gaps in the following positions: 2984. However, reference sequence is expected to be a high-quality, complete, and unambiguous as much as possible. Gaps are not allowed.

Location:
   packages/nextclade/src/run/validate_ref_seq.rs:31

```


packages/nextclade/src/run/validate_ref_seq.rs
```diff
@@ -0,0 +1,66 @@
...
+    return make_error!("Sequence is empty");
+  }
+  Ok(())
+}
+
+fn seq_check_gaps(seq: &[Nuc]) -> Result<(), Report> {
+  let ref_gap_ranges: Vec<NucRange> = find_letter_ranges(seq, Nuc::Gap);
+  if !ref_gap_ranges.is_empty() {
+    let ranges_formatted = ref_gap_ranges.into_iter().map(|r| r.range.to_string()).join(",");
+    return make_error!("Sequence contains gaps in the following positions: {ranges_formatted}. However, reference sequence is expected to be a high-quality, complete, and unambiguous as much as possible. Gaps are not allowed.");
```
victorlin, line 31: Rewording suggestion:

```suggestion
    return make_error!("Sequence contains gaps in the following positions: {ranges_formatted}. Gaps are not allowed: reference sequence is expected to be high-quality, complete, and unambiguous.");
```

Issue #1620, opened 2025-05-15
Nextclade outputs ANSI escape codes when redirected to a file


genehack:
When redirecting output to a file, Nextclade still emits ANSI color codes, resulting in file contents such as:

```
[2m2025-05-15 23:00:02.536[0m [2m[[0m[2m[36mI[39m[0m[2m][0m [2mhttp_client.rs:122:[0m HTTP 'GET' request to 'https://data.clades.nextstrain.org/v3/index.json'
[2m2025-05-15 23:00:02.995[0m [2m[[0m[2m[36mI[39m[0m[2m][0m [2mhttp_client.rs:122:[0m HTTP 'GET' request to 'https://data.clades.nextstrain.org/v3/nextstrain/yellow-fever/prM-E/2025-03-26--11-47-13Z/dataset.zip'
```

It would be better if Nextclade only emitted ANSI color codes when it detects it is writing to a terminal.

PR #1625, opened 2025-05-20
fix(cli): ensure proper CLI color mode handling

ivan-aksamentov:
Resolves https://github.com/nextstrain/nextclade/issues/1620

Unify color output configuration, respect user and environment preferences. Improve portability and correctness by migrating from `owo-colors` to `console`.

- Replace `owo-colors` with [`console`](https://docs.rs/console) for global color control and better TTY integration (NOTE: `color_eyre` still uses `owo-colors` underneath)
- Fix `color_eyre` (errors, panics and stack traces handler) init to take into account global color config
- Use [`console::set_colors_enabled()`](https://docs.rs/console/latest/console/fn.set_colors_enabled.html) internally to enable or disable color output globally
- Implement color mode resolution in order of preference:
  - CLI flags (`--color`, `--no-color`)  - last occurrence wins
  - Environment variables: `CLICOLOR_FORCE`, `NO_COLOR`, and `COLOR`
  - Whether we are writing to TTY or not
- move most of global init code from `ctor` (static initialization) to after CLI args parser, to make sure runtime config is taken into account
- `console` crate does not support hex colors, so had to approximate hex to ANSI 265 color conversion
- Ensure compatibility with `wasm32` build for web app, where we do use colored logging, but should not use `std::env` and TTY checks

The new interface is as follows (see `nextclade --help` or reference docs):

```
--color <COLOR>
    Control when to use colored output [env: COLOR, NO_COLOR, CLICOLOR_FORCE]
    
    Color output control follows this precedence (highest to lowest):
    
    1. Command-line flags:
    
    - `--color`
    
    - `--no-color`
    
    2. Environment variables:
    
    - `CLICOLOR_FORCE=1` - force color
    
    - `NO_COLOR` (set) - disable color
    
    - `COLOR=always|auto|never`
    
    If both `--color` and `--no-color` are provided, the one specified last takes effect.

    Possible values:
    - auto:   Automatically enable colors if output is a TTY (default)
    - always: Always enable colors
    - never:  Never enable colors

--no-color
    Disable colored output (shorthand for --color=never)
    
    This overrides all related color environment variables (`CLICOLOR_FORCE`, `NO_COLOR`, `COLOR`)
    
    If both `--color` and `--no-color` are provided, the one specified last takes effect.
```

### DIFF
--- a/docs/user/nextclade-cli/reference.md
+++ b/docs/user/nextclade-cli/reference.md
@@ -54,6 +54,39 @@ For short help type: `nextclade -h`, for extended help type: `nextclade --help`.
 * `--silent` — Disable all console output. Same as `--verbosity=off`
 * `-v`, `--verbose` — Make console output more verbose. Add multiple occurrences to increase verbosity further
 * `-q`, `--quiet` — Make console output more quiet. Add multiple occurrences to make output even more quiet
+* `--color <COLOR>` — Control when to use colored output [env: COLOR, NO_COLOR, CLICOLOR_FORCE]
+
+   Color output control follows this precedence (highest to lowest):
+
+   1. Command-line flags:
+
+   - `--color`
+
+   - `--no-color`
+
+   2. Environment variables:
+
+   - `CLICOLOR_FORCE=1` - force color
+
+   - `NO_COLOR` (set) - disable color
+
+   - `COLOR=always|auto|never`
+
+   If both `--color` and `--no-color` are provided, the one specified last takes effect.
+
+  Possible values:
+  - `auto`:
+    Automatically enable colors if output is a TTY (default)
+  - `always`:
+    Always enable colors
+  - `never`:
+    Never enable colors
+
+* `--no-color` — Disable colored output (shorthand for --color=never)
+
+   This overrides all related color environment variables (`CLICOLOR_FORCE`, `NO_COLOR`, `COLOR`)
+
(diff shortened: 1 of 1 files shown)


genehack:
> > nextclade --help --color=never still gives colorized output,
[snip explanation]

> For now the choice is:
> 
>     * remove `--color` and `--no-color` as buggy and leave only env vars
> 
>     * keep the args and submit a known issue until there's some obvious way how to handle that
> 
> What do you think?
> 
> Which one you'd prefer as a user?

I think _not_ having `--color/--no-color` and being forced into only envvar usage would annoy me more than the alternative. :grin: Second option is fine by me. 

> > running nextclade --color=never results in (colorized) output saying:
[snip]

> I am not sure whether that's a bug or a feature. If we consider this a bug, it can be either in `clap` or in how I use it. I made an issue #1627 

👍 

> > not clearly grasping the precedence
> 
> I tried to implement the following:
> 
>     * output target (TTY or not) - is the lowest priority factor, the "sane default" that works well without any action
> 
>     * next, env vars  - something that needs to be set intentionally, but is not specific to Nextclade, comes from global environment - they override the default
> 
>     * then the CLI args  - something that's very specific to Nextclade and very intentional - they override everything and, if multiple args present, the args that comes last wins
> 
> Does it make sense?

yeah, that makes sense, and reviewing the various things I tried, I think that matches the behavior I saw (modulo the first issue, above).

+1 from me on merging.

Issue #1631, opened 2025-05-27
Nextclade Web: Unexpected error occurred


joverlee521:
I was testing a [new dataset](https://clades.nextstrain.org/?dataset-url=https://github.com/theiagen/rabies/tree/main/nextclade/minimal-dataset) with @kimandrews and ran into a strange behavior. It's unclear if this is just an improperly set up dataset or if there's a bug in Nextclade.

The results page does not list any of the samples even though I see the analysis progress bar completed

![Image](https://github.com/user-attachments/assets/cbd5a0b5-cc49-4067-b26b-9d29a6483c3a)

The tree page seems fine with all of the samples placed on the reference tree. However, the export page displays the "An unexpected error has occurred" message with the following:

```

Error message: Error: Internal Error: Dataset not found: 'https://raw.githubusercontent.com/theiagen/rabies/main/nextclade/minimal-dataset'. This is an internal issue, likely due to a programming mistake. Please report it to developers!

Nextclade version 3.14.0 (commit: 6089875, branch: release)

Memory available: 3586 MBytes

User agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36

Browser details: {"browser":{"name":"Chrome","version":"136.0.0.0"},"os":{"name":"macOS","version":"10.15.7","versionName":"Catalina"},"platform":{"type":"desktop","vendor":"Apple"},"engine":{"name":"Blink"}}

Call stack:

Error: Internal Error: Dataset not found: 'https://raw.githubusercontent.com/theiagen/rabies/main/nextclade/minimal-dataset'. This is an internal issue, likely due to a programming mistake. Please report it to developers!
    at cc (https://clades.nextstrain.org/_next/static/chunks/pages/export-f7841e27d85ffff9.js:1:61596)
    at fW (https://clades.nextstrain.org/_next/static/chunks/framework-5774a691a4029c28.js:1:61454)
    at _ (https://clades.nextstrain.org/_next/static/chunks/framework-5774a691a4029c28.js:1:119698)
    at h5 (https://clades.nextstrain.org/_next/static/chunks/framework-5774a691a4029c28.js:1:107754)
    at h3 (https://clades.nextstrain.org/_next/static/chunks/framework-5774a691a4029c28.js:1:107682)
[...]

PR #1632, opened 2025-05-28
fix(web): crash when using single dataset from url

ivan-aksamentov:
Resolves: #1631

Nextclade Web crashed due to a race between single-dataset initialization from url params and periodic background re-fetch of datasets (background dataset updates to catch the new releases). The updates coming from the default dataset server would overwrite the custom dataset in the app state.

The first periodic fetch would happen right after the actual custom dataset was fetched, as main page starts to render, but after app component is done initialization. So manual runs would crash because there is enough time for the update to kick in, but automatic runs with `?input-fasta` would not, because the main page is never rendered and skipped to the results page.

Additional nuissance is that adding sequences manually would trigger dataset suggestion algo and it would fail, because the custom dataset is obviously not present in the minimizer index and there are no matches possible. This manifested in an error alert to the users, even though running suggestions does not make sense in the first place.

Here I try to mitigate this situation, by setting  current dataset server url state atom to `undefined` whenever a custom dataset is requested with `?dataset-url=` or `?dataset-json-url=`. This way is nowhere to fetch the dataset index or data or minimizer index from, so:
 - periodic background dataset fetch knows to disable itself
 - dataset suggestion algo is skipped

This has a defect - once launched with a custom dataset, the app enters a state where there is only this dataset and nothing else. But I think it was the same in previous releases too. The workaround is to refresh the browser page (or open the app without url params). I'd like to address this in a separate PR: `<insert issue number here>`

Overall, the initialization code has become a bit of a spaghetti at this point, and we need to refactor and disentangle the init and app state at some point in order prevent these kind of edge cases from happening.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,6 +4,8 @@
 # produce it. Refer to osxcross documentation for more info.
 # This is only needed if you are on Linux and want to produce binaries for macOS.
 OSXCROSS_URL=http://example.com/osxcross/osxcross.tar.xz
+MUSL_CC_AARCH64_URL=http://example.com/musl.cc/11/x86_64-linux-musl/aarch64-linux-musl-cross.tgz
+MUSL_CC_X86_64_URL=http://example.com/musl.cc/11/x86_64-linux-musl/x86_64-linux-musl-cross.tgz
 
 FULL_DOMAIN=autodetect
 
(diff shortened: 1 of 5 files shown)


joverlee521:
Thanks for the quick fix @ivan-aksamentov! The custom dataset via `?dataset-url=` works as expected for me in the preview app.

Issue #1265, opened 2023-10-03
Web: Grey scale coloring for region/country/divison if scale not predefined in reference tree


joverlee521:
It's not clear to me if this should be considered a bug or user-error but at least wanted to document the behavior here. 

If the reference tree for a Nextclade dataset does not pre-define the color scale for region, country, or division fields, then coloring becomes grey scale due to the injection of the "Unknown" value scale

https://github.com/nextstrain/nextclade/blob/aaac7ce891ce1a83b6554097a0a075004e770282/packages_rs/nextclade/src/tree/tree_preprocess.rs#L282-L283

This was first reported in Nextstrain office hours on 2023-09-21 by an external user who built their own [Nextclade datasets](https://clades.nextstrain.org/?dataset-url=https://github.com/mazeller/nextclade_test/tree/main/prrsv_zhang_2023), but can be observed in the "official" Influenza A H1N1pdm HA dataset which does not define the color scale for "country":
![Screen Shot 2023-10-03 at 2 49 42 PM](https://github.com/nextstrain/nextclade/assets/40774762/bf58d036-1b30-4e9f-ab53-b4c0bf891a04)

joverlee521:
I was wondering why the "Unknown" value wasn't automatically ignored by Auspice and found this was done on purpose!

https://github.com/nextstrain/nextclade/blob/ca54ba3b13ccb8ac8a4aad0e8b37e0d583395598/packages_rs/nextclade/src/tree/tree.rs#L21-L23

ivan-aksamentov:
Thanks Jover!

Since the beginning, we've been trying to carefully ignore the country and other values, because Nextclade inputs and, therefore, newly attached tree nodes simply don't have this information.

My understanding is that in our own datasets we don't have these fields, and so these colorings do not even appear in the dropdown.

I had to add the handling of unknown values long time ago for the reasons I don't remember clearly, but I found this Slack thread:
https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1596559922285700?thread_ts=1596488014.283400&cid=C7SDVPBLZ

But maybe it needs to be revisited.


You know more about this things. What can you recommend?

joverlee521:
Ah, the [grey scale is the intended behavior in Auspice for values that are not defined in the color scale](https://github.com/nextstrain/auspice/blob/8377c4c47003468dab5b34cc02ab2d06eb952ca1/src/util/colorScale.js#L139-L147). 

[...]

PR #1646, opened 2025-06-05
fix: country colorings in output tree json

ivan-aksamentov:
Resolves https://github.com/nextstrain/nextclade/issues/1265

This removes the special value "Unknown " (with a space) attached to the "region", "country" and "division" node attributes for new nodes, as well as a color scale for this value.

This way:
- the fields no longer appear in the json (for newly attached nodes)
- the "Region": "Unknown" etc. lines no longer show up in the popup in Auspice interface when clicking on a new node
- colors are displayed correctly, with new nodes in grey

### DIFF
--- a/packages/nextclade/src/tree/tree.rs
+++ b/packages/nextclade/src/tree/tree.rs
@@ -15,16 +15,12 @@ use eyre::{eyre, Report, WrapErr};
 use log::warn;
 use schemars::JsonSchema;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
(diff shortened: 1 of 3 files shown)


joverlee521 (approved):
Thanks for the fix @ivan-aksamentov! Tested in the preview app with the [H1N1pdm dataset](https://nextstrain--nextclade--pr-1646.previews.neherlab.click/?dataset-name=nextstrain/flu/h1n1pdm/ha/CY121680) and it works as I expect.

PR #1648, opened 2025-06-09
chore: switch off LTO for profiling profile

corneliusroemer:
With lto on, I didn't get any insight into what was going on inside `sort`, everything was apparently inlined by LTO.

With lto off, I now see what's going on.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -10,6 +10,7 @@ lto = "off"
 inherits = "release"
 debug = 1
 strip = false
+lto = "off"
 
 # Optimize dependencies even in debug mode
 [profile.dev.package."*"]


corneliusroemer:
You just don't see useful things in stacktrace with lto on - maybe we could have 2 profiles, one with lto one without.

I also noticed we might actually get faster performance with thin lto rather than fat!

corneliusroemer:
> But in all cases it's important to make sure that we optimize for what user is getting and not just chasing pink elephants that don't exist in prod (e.g. a useful call stack - it's just not there anymore)

I'm not chasing pink elephants :) - I just use profiling to find likely culprits, and test with release to confirm the speedups work.

Issue #1647, opened 2025-06-07
Suggestion crashes with OOM from around ~3k HA seqs (6MB)


corneliusroemer:
While it's fine to run the suggestion algorithm on around a thousand sequences, it seems to become unusable for only around ~3000 short flu HA sequences.

Repro:
1. Download these sequences: https://api.loculus.genspectrum.org/influenza-a/sample/unalignedNucleotideSequences/seg1?downloadAsFile=true&downloadFileBasename=influenza-a_nuc-seg1_2025-06-07T1034&dataFormat=fasta&versionStatus=LATEST_VERSION&isRevocation=false&country=Australia&length_seg1From=2000
2. Drop them into Nextclade and run suggest (take the time it takes)
3. Drop the file a second time into Nextclade (to double the number of sequences). Click re-suggest.

For flu on my machine, 500 seqs take around 2 seconds to suggest (with all wasm/minimizer cached). 1000 seqs take 4 seconds. 2000 seqs around 8 seconds. But from around 3k I get OOM crashes (after around 40 seconds).

PR #1649, opened 2025-06-09
perf(sort): use lookup table to speed up sequence conversion to bitpacked representation

corneliusroemer:
Looking into sort performance (see https://github.com/nextstrain/nextclade/issues/1647), I noticed that the various string containment operations where performance limiting.

Instead of doing multiple string containment call, this PR uses a compile time lookup table to turn a sequence into a bitpacked representation.

A quick benchmark on 240MB of SC2 sequences show this reduces core sort algorithm runtime (excluding minimizer download) by 42%.

### DIFF
--- a/packages/nextclade/src/sort/minimizer_search.rs
+++ b/packages/nextclade/src/sort/minimizer_search.rs
@@ -285,7 +285,7 @@ pub fn calculate_minimizer_hits(
 
     seq_str.windows(k)
         .map(|kmer| get_hash(kmer, params))
-        .filter(|&mhash| mhash < cutoff && seen.insert(mhash))
+        .filter(|&mhash| mhash < cutoff && seen.insert(mhash)) // Faster than .unique()
         .fold(vec![0; n_refs], |mut acc, m| {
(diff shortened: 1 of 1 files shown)


packages/nextclade/src/sort/minimizer_search.rs
```diff
@@ -243,32 +243,40 @@ const fn invertible_hash(x: u64) -> u64 {
...
+  for (i, &nuc) in kmer.iter().enumerate() {
     if i % 3 == 2 {
-      continue; // skip every third nucleotide to pick up conserved patterns
-    }
-
-    if !"ACGT".contains(nuc) {
-      return cutoff + 1; // break out of loop, return hash above cutoff
-    }
-
-    // A=11=3, C=10=2, G=00=0, T=01=1
```
corneliusroemer, line 263: This comment was actually wrong - T and C should have been interchanged.

corneliusroemer: That's why the initial preview finds no matches (I used the comment rather than the code to implement the lookup table)

Issue #1660, opened 2025-07-02
attributes contain URL-encoded strings


fhennig:
We noticed that `attributes` from nextclade contain URL-encoded semicolons. These are URL-encoded in the `gff3` files, but we'd expect that nextclade does the decoding already.

https://github.com/loculus-project/loculus/pull/4490#discussion_r2179279881

anna-parker:
Hi Ivan!

We are unfortunately using the JSON output (this seemed easiest as we use the annotations to annotate embl flat files for ENA which are slightly different than genbank files and the team wanted us to be able to store annotations in a JSON format to use for different file types in future). I was aware that there might be changes to the JSON and am happy to adapt the code to the new changes. 

Concerning the url encoding. I was also uncertain initially if it would be best to use a custom -built encoder for the specific characters that they mention. However I think the encoder you chose should cover all the correct cases - at least for decoding the input attributes. The only issue I could see is that the encoder might end up encoding too many characters in the output gff3 - but we can probably handle that case if it comes up @corneliusroemer what is your opinion.

PR #1661, opened 2025-07-03
fix: ensure gff fields and attributes are encoded and decoded

ivan-aksamentov:
Resolves: https://github.com/nextstrain/nextclade/issues/1660

Here I apply URL-decoding on GFF read and URL-encoding on GFF write to all attribute values.

Quote from [GFF3 spec](https://github.com/The-Sequence-Ontology/Specifications/blob/fe73505276dd324bf6a55773f3413fe2bed47af4/gff3.md#:~:text=URL%20escaping%20rules%20are%20used%20for%20tags%20or%20values%20containing%20the%20following%20characters%3A%20%22%2C%3D%3B%22.%20Spaces%20are%20allowed%20in%20this%20field%2C%20but%20tabs%20must%20be%20replaced%20with%20the%20%2509%20URL%20escape.):

> Column 9: "attributes"
> [..] URL escaping rules are used for tags or values containing the following characters: ",=;". Spaces are allowed in this field, but tabs must be replaced with the %09 URL escape [..]

### DIFF
--- a/packages/nextclade/src/features/feature.rs
+++ b/packages/nextclade/src/features/feature.rs
@@ -1,13 +1,12 @@
 use crate::coord::range::NucRefGlobalRange;
 use crate::features::feature_type::shorten_feature_type;
 use crate::gene::gene::GeneStrand;
-use crate::io::gff3_encoding::gff_decode_attribute;
+use crate::io::gff3_encoding::gff_decode_non_attribute;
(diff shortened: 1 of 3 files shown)


anna-parker (approved):
Thanks so much Ivan! I just tested this branch locally on the example I sent you and this works beautifully! I'm not a rust expert but the code also looks great - thanks again for looking into and fixing this!

PR #1700, opened 2025-10-17
feat: add "recombinant" qc metric

### LINKED ISSUE #1699: QC label for recombinant sequences
I would like to propose the addition of a QC label specifically for "recombinant" sequences (e.g., an "R" label). This feature should:

- Assess the number of (private) mutations and reversions in a sequence.
- Highlight and label the sequence as recombinant when the count exceeds a configurable threshold.

This will help users quickly identify sequences that are likely recombinant and may require special attention in downstream analyses.

**Details:**
- The threshold for labeling should be adjustable/configurable.
- The label should be visible in the QC results and ideally in the output reports.
- Optionally, provide a summary or explanation for sequences that receive the label (e.g., why it was labeled recombinant).

Let me know if you need further details or use case examples 😊

ivan-aksamentov:
Resolves #1699

Example data is in: https://github.com/nextstrain/nextclade_data/pull/367

Test web app with: https://nextstrain--nextclade--pr-1700.previews.neherlab.click/?dataset-server=gh:@feat/qc-recombinants@&dataset-name=sars-cov-2&input-fasta=example

### DIFF
--- a/packages/nextclade/src/qc/qc_rule_recombinants.rs
+++ b/packages/nextclade/src/qc/qc_rule_recombinants.rs
@@ -1,5 +1,6 @@
 use crate::analyze::find_private_nuc_mutations::PrivateNucMutations;
 use crate::qc::qc_config::QcRulesConfigRecombinants;
+use crate::qc::qc_rule_snp_clusters::QcResultSnpClusters;
 use crate::qc::qc_run::{QcRule, QcStatus};
 use num::traits::clamp_min;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


packages/nextclade/src/qc/qc_config.rs
```diff
@@ -233,6 +233,27 @@ impl QcRulesConfigStopCodons {
...
+
+impl QcRulesConfigRecombinants {
+  pub const fn example() -> Self {
+    Self {
+      enabled: true,
+      mutations_threshold: 20,
+      score_weight: OrderedFloat(100.0),
+    }
+  }
+}
```
ivan-aksamentov: Configurable variables can be adjusted here (maps to fields in `qc.recombinants` of `pathogen.json`). They are passed to the function which calculates the metric.

PR #1765, opened 2026-04-28
fix: treat deletions as undetermined in nearest node distance

ivan-aksamentov:
- Related: https://neherlab.slack.com/archives/C015PFP5V44/p1777319589953279

Sequences with large internal deletions are placed near the root of the reference tree instead of near their true closest relatives. The nearest node distance metric counts node mutations at deleted positions as mismatches (the query "has reference" there), making the root -- which has zero mutations -- appear closest. Replacing gaps with N produces correct placement because N positions are already excluded from the distance as undetermined.

This PR adds query deletion ranges as a separate parameter to the distance calculation and counts node mutations at deleted positions as undetermined. The fix is limited to placement; private mutation calling already handles deletions correctly via a separate code path.

### Work items

- [x] Add `qry_deletions: &[NucDelRange]` parameter to [`graph_find_nearest_nodes`](https://github.com/nextstrain/nextclade/blob/3004cda11b044866131731610246752533f9f646/packages/nextclade/src/tree/tree_find_nearest_node.rs#L21) and [`tree_calculate_node_distance`](https://github.com/nextstrain/nextclade/blob/3004cda11b044866131731610246752533f9f646/packages/nextclade/src/tree/tree_find_nearest_node.rs#L77)
- [x] Count node mutations at deleted positions as undetermined in the [distance loop](https://github.com/nextstrain/nextclade/blob/3004cda11b044866131731610246752533f9f646/packages/nextclade/src/tree/tree_find_nearest_node.rs#L121-L126)
- [x] Pass `&deletions` from [`nextclade_run_one`](https://github.com/nextstrain/nextclade/blob/3004cda11b044866131731610246752533f9f646/packages/nextclade/src/run/nextclade_run_one.rs#L330)
- [x] Add unit test verifying deletion-covered node mutations reduce distance
- [x] Update existing tests for new parameter signature

### DIFF
--- a/packages/nextclade/src/tree/tree_find_nearest_node.rs
+++ b/packages/nextclade/src/tree/tree_find_nearest_node.rs
@@ -1,6 +1,7 @@
 use crate::alphabet::nuc::Nuc;
 use crate::analyze::is_sequenced::is_nuc_sequenced;
 use crate::analyze::letter_ranges::NucRange;
+use crate::analyze::nuc_del::NucDelRange;
 use crate::analyze::nuc_sub::NucSub;
 use crate::coord::range::NucRefGlobalRange;
 use crate::graph::node::GraphNodeKey;
@@ -21,6 +22,7 @@ pub fn graph_find_nearest_nodes(
   graph: &AuspiceGraph,
   qry_nuc_subs: &[NucSub],
   qry_missing: &[NucRange],
+  qry_deletions: &[NucDelRange],
   aln_range: &NucRefGlobalRange,
 ) -> Result<Vec<TreePlacementInfo>, Report> {
   let masked_ranges = graph.data.meta.placement_mask_ranges();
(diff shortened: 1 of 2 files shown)


packages/nextclade/src/tree/tree_find_nearest_node.rs
```diff
@@ -105,11 +115,13 @@ fn tree_calculate_node_distance(
...
   // for these we can't tell whether the node agrees with seq
   let mut undetermined_sites = 0_i64;
   for pos in node.tmp.substitutions.keys() {
-    if !is_nuc_sequenced(*pos, &masked_qry_missing, aln_range) {
+    if !is_nuc_sequenced(*pos, &masked_qry_missing, aln_range)
+      || qry_deletions.iter().any(|del| del.range().contains(*pos))
+    {
       undetermined_sites += 1;
     }
   }
```
ivan-aksamentov, line 127: Dels are new passed to distance metric calculation and counted same as unknowns or unsequenced. But I imagine it can also be treat differently if needed.

packages/nextclade/src/tree/tree_find_nearest_node.rs
```diff
@@ -358,10 +434,51 @@ mod tests {
...
+      &qry_missing,
+      &qry_deletions,
+      &aln_range,
+      &masked_ranges,
+    );
+    assert_eq!(distance_with_del, 2);
+
+    Ok(())
+  }
 }
```
ivan-aksamentov, line 484: Tests that adding a del reduces distance metric

rneher:
Thanks! this places these sequences as expected and is I think a universal improvement. One could add `+1` to the distance for every deletion that is not in the attachment node, but deletions are represented differently on the auspice tree nodes and ignoring distance by deletion is in line with what phylo tools generally do... So it is probably not worth the trouble.

